PR #28, opened 2017-05-26
Fixed #25 NoRTC overflow and bugs

### LINKED ISSUE #25: Fix overflow in UpdateNoRTC for NO_RTC build
> What I need to do here is mask off the top bits and check if the current hi-days is %11111 (i.e. just before it would overflow). When that happens there's a few possibilities here:
> 
> * Reset the whole RTC to zero
> * Go back a week (so the last week repeats itself indefinitely)
> * Repeat the last day indefinitely
> 
> NOTE: 13 bits for days allows for 8'192 in-game days -- over 22 in-game years -- which, given that a day in-game takes 4 real hours, would require the player to play for 3.75 continuous real years to overflow.

Kroc:
The last in-game year will repeat.
The uncommented block was causing problems and doesn't appear needed in the end, but further testing will tell.

### DIFF
--- a/home/game_time.asm
+++ b/home/game_time.asm
@@ -150,23 +150,24 @@ UpdateNoRTC::
 	ld hl, wNoRTCSeconds
 
 ; +1 second
-	ld a, b
-	inc [hl]
-	sub [hl]
-	ret nz
-	ld [hld], a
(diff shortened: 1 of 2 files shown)


roukaour:
Thank you very much!

PR #29, opened 2017-05-28
Fix wild Pokémon generation

FredrIQ:


FredrIQ:
The last commit should address your issues.

PR #116, opened 2017-08-04
Fix NoRTC patch behaviour

### LINKED ISSUE #113: No-RTC clock gets stuck in a loop
If you start a game at Sunday 20:00 and let time pass, it will reach Monday 19:59 correctly, but then loops back to Sunday 20:00 instead of Monday 20:00.

image

Edit: Reset the clock by pressing Down+B at the title screen. Sunday 0:00 turns into Saturday at 0:30. Thursday 10:07 turns into Wednesday at 10:30.

Kroc:
Fixes #113


roukaour:
Hey, thanks a lot! Your changes included a lot of repeated comments and whitespace edits, so I manually copied the relevant ones into commit b027ecc5df10517eedefc8d34220dcb4032435ce. (Which also let me study how your patch works.)

It seems to work fine in RTC and no-RTC versions. A slight issue: when I save the game, rebuild the ROM in the opposite version, and reload, the time has changed. For instance, Sunday 2:42 PM in no-RTC became Sunday 12:30 in RTC.

This could just be due to me miscopying something from your PR. But would you mind investigating it? (I would, but I have all of the map scripts to go through and update with new PERSONTYPEs.) If you can't, it's no big deal; I don't expect people to switch between versions often, and they can reset the time with Down+B if needed.

Kroc:
That behaviour is normal because the RTC is being read from different places in the different versions! When you save with NoRTC, the clock is being written to the SRAM. That time is not being read when a real RTC is present -- you're getting a different time because the duration of a real RTC will be different from the 6x speed being manually computed and stored.

roukaour:
Heh. That makes sense then.

PR #268, opened 2019-09-17
Psychic only has 10 PP

SadisticMystic:


Rangi42:
Thank you for finding and fixing this!

Issue #264, opened 2019-08-25
The Town Maps in Pokémon Centers show the player sprite as solid white


Rangi42:
![image](https://user-images.githubusercontent.com/35663410/63656773-df646100-c766-11e9-9f49-6d74f22679ec.png)

PR #269, opened 2019-09-28
Fix issue #264: white sprites in Pokecenter town map

itsdarsh:
Sprite palettes were never initialized after FadeToWhite set all wUnknOBPals values to $FF


Rangi42:
Nice, thanks for fixing this!

Issue #256, opened 2019-03-31
Various mechanical errors


SadisticMystic:
I play on a 3-beta faithful build, the most recent such patch linked in the Discord. Along the way, I've noticed several mechanical interactions that work differently than I should expect from the precedent set in later official games. Some of them might be harder to implement in the established control flow of the existing codebase than others, and I may add to this list as I notice more oddities. 

- [x] If there's an AI manual switch to a Pokemon with Trace, Trace doesn't activate there in PC. Instead, it waits for the next player switch, and announces itself at that point to take the new ability you switched to. It's supposed to fire immediately no matter how the Tracer was brought in.  
- [x] Ghost-types are supposed to have an automatic Run Away effect as far as determining if they can get away from a wild battle safely  
- [x] If U-turn or Volt Switch scores a KO, experience is supposed to be handed out before you bring a new Pokemon in, and the newcomer is not supposed to be part of the experience split unless they were already tagged as having a previous stint in the active slot    
- [ ] In modern games, when trainer AI makes a manual switch, the Pokemon that were tagged as having spent time in battle against the one that switched out should be restored the next time that Pokemon comes back. PC does not honor that, so if you're trying to do switch-training on a weak piggyback, an opponent's switch completely destroys any possible benefit you can get from that, unless you somehow find away to get the piggyback in safely for one more turn.  
- [x] Acrobatics is exactly the opposite of how it should be: it gets the power boost if you are holding an item, not otherwise  
- [x] Gyro Ball appears to have a problem computing its power, and is always dealing 1 damage even when it should be much more powerful than that  
- [x] Based on extensive observation of the opponent's switching patterns in several battles that dragged out for a while, I have a high degree of confidence that trainer AI's decisions on whether to switch at all, and if so what to bring out, are being delayed until they have received undue foreknowledge of what the player has committed to that very same turn as far as their own switch  
- [x] It is possible for an opposing trainer to use Pursuit on the player's switch and still go on to use an item in the very same turn, if they were planning on using an item (possibly related: in the case where this behavior was observed, Pursuit hit into Gengar and tripped Cursed Body, so that might have caused them to forget they had already used the move in the middle of the turn)  
- [x] When one player goes first with U-turn or Volt Switch and the other had selected Pursuit, Pursuit is supposed to happen after U-turn's damage (assuming it didn't KO) but before the switch, and be double power; current PC behavior causes Pursuit to hit after the switch in this case  
[...]

PR #270, opened 2019-09-29
Refactor Pursuit/switch and fix U-turn/drain move issues

itsdarsh:
This fix addresses part of #256

Drain moves base their recovery on the floor of 1/2 of wCurDamage (or 1, if only 1 damage was dealt). In base Crystal, there is code in the subtract HP function that sets wCurDamage to the original value of wPlayerMonHP/wEnemyMonHP before changing those values if the Pokemon takes more damage than it has remaining HP (in other words, set wCurDamage to what the Pokemon's HP was before it fainted). This fix re-implements that code, which will also make sure Pokemon take correct damage on recoil damage (which also bases its recoil damage on wCurDamage). This adds 32 cycles to each damage calculation.


aaaaaa123456789:
This will not compile. Did you test it?

itsdarsh:
Sorry! I typed it in by hand but I really should've copied and pasted it in. It should've been `ld [wCurDamage+1], a` at line 1926 and `ld [wCurDamage], a` at line 1929 but I forgot to put brackets in appropriately. It should compile now.

FredrIQ:
If A uses U-Turn first, B responds with immediate Pursuit before the switch-out and is knocked out with Rocky Helmet, does things still work as they should?

itsdarsh:
Probably not, I forgot all about Rocky Helmet fainting a Pursuit user (or similarly that user dying of Life Orb recoil). There's also a visual bug where the fainted opponent's HP bar will reload after you select a Pokemon if you knock out the enemy mon with U-turn. I'll definitely get the faint after Pursuit issue working, since that would also affect a switch out resulting from Teleport during a trainer battle (which calls ContinueToSwitchOut).

itsdarsh:
PursuitSwitch no longer copies wLastPlayerMon to wCurBattleMon to run, removing a TODO

PR #285, opened 2019-10-27
three minor bug fixes

### LINKED ISSUE #281: Game crashes when starting a battle with no Pokemon
Not really a HUGE issue for obvious reasons, but in the originals, it would just fade to white and back, skipping the battle. It just crashes in PC.

### LINKED ISSUE #282: Extra character drawn on the "Cancel" line in pack if last Key Item is registered
If the last Key Item in the Key Item pocket is registered, it's "registered" symbol is drawn on the next line in front of "Cancel". Happens with the star, and all of the directional symbols, too.
star

itsdarsh:
Fixes #281 - the crash occurs in PostBattleTasks, so putting in an appropriate check for an empty party fixes that

Fixes #282 - uses a different set of flags for the Key Items pocket to change how value $FF (Cancel) is handled. Normally, when a $FF item value is found, the general scrolling menu function just skips any special functions associated with the item menu and directly prints the cancel string in the appropriate slot. This is normally fine, but Key Items have a special property in PlaceMenuItemName where a symbol is drawn if they're registered, and this symbol is never actually reset in the tilemap if you scroll down from a registered item to Cancel. Thus, a call to PlaceMenuItemName is needed to blank that tile for Cancel, and a check is needed to dodge a call to GetItemName (as the item name for $FF is "?@", not "Cancel@", and I wanted to preserve that).

Also fixes an issue with the Johto map (flying, pokegear, pokedex, and town map) where the tile at x = 07, y = 01 was flipped in both the x and y directions for some reason. Changed the offending value from $C3 to $13.


Rangi42:
Excellent documentation of the bugs and fixes, thank you!

PR #289, opened 2019-11-10
set of bug fixes, including switch/predict bug

### LINKED ISSUE #277: Potential crash on evolving Pikachu
If Pikachu's form isn't properly loaded into memory (i.e. by looking at its stats page), evolving Pikachu will lead to the game calling an invalid front pic when evolving it, leading to a crash.

This can be fixed just by fetching the form party parameter in EvoStoneFunction, but I believe there's a more efficient way to do this. I think item_effects.asm can be optimized a bit more rather than just adding another nine bytes to the file, so I'll just leave this bug report here for posterity's sake.

### LINKED ISSUE #287: Switch/Predict switching after enemy Pokemon faint in Trainer Battle is broken
During a trainer battle, if the battle style is set to Switch or Predict, attempting to switch after fainting a Trainer Pokemon does not properly work.  Either the game glitches out and acts as if a glitched Pokemon is fainted upon sending it out, forcing you back to picking another Pokemon, or it sends in the wrong Pokemon altogether.

itsdarsh:
Fixes #287 - this one was entirely my bad, when I removed a ton of code from PursuitSwitch to remove the reliance on wLastPlayerMon, I didn't realize that I'd unwittingly removed a bit of code that properly sets which Pokemon we switch into. Loading wCurBattleMon back into wCurPartyMon (since wCurBattleMon is exclusive to the player mon) before calling PlayerSwitch would seem to fix that problem.

Fixes #277 and refactors item_effects.asm - item_effects.asm honestly needed to be cleaned up a bit, and based on my testing the modifications I've made have preserved the exact functionality of the original file, just with a bit more efficiency. The issue with crashing upon evolution is better resolved just by setting wCurForm during the evolution script, since it's only otherwise needed for rare candies and evolution stones and isn't otherwise necessary for any item effects, and is more efficient to insert in evolve.asm rather than fetching it twice in item_effects.asm

Fixes a minor visual bug with the party menu after using an item in battle - because the tilemap wasn't cleared after using an item (such as a potion) on a pokemon in battle, there is a brief moment where the party menu opens up with the palettes of the enemy pokemon before we see the normal battle screen, so adding a conditional call to ClearTileMap addresses that.

Modifies makefile - now, `make` and `make faithful` will create roms with different filenames, so you can build both the faithful and nonfaithful versions without needing to mess around with directories. Same for debug, and arguably we can just extrapolate this to the other options as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,16 +85,12 @@ crystal: FILLER = 0x00
 crystal: ROM_NAME = $(NAME)-$(VERSION)
 crystal: $(NAME)-$(VERSION).gbc
 
-faithful: ROM_NAME = $(NAME)-$(VERSION)-faithful
-faithful: $(NAME)-$(VERSION)-faithful.gbc
-
+faithful: crystal
 nortc: crystal
 monochrome: crystal
 noir: crystal
 hgss: crystal
-
-debug: ROM_NAME = $(NAME)-$(VERSION)-debug
-debug: $(NAME)-$(VERSION)-debug.gbc
+debug: crystal
 
 bankfree: FILLER = 0xff
 bankfree: ROM_NAME = $(NAME)-$(VERSION)-0xff
--- a/Makefile
+++ b/Makefile
@@ -85,12 +85,16 @@ crystal: FILLER = 0x00
 crystal: ROM_NAME = $(NAME)-$(VERSION)
 crystal: $(NAME)-$(VERSION).gbc
 
-faithful: crystal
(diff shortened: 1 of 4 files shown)


Rangi42:
Nice bugfixing. :) Please undo the Makefile change though; it looks like it wouldn't support arbitrary combinations of flags, e.g. `make faithful debug`.

itsdarsh:
Sure, they're undone. I forgot flags could be combined, sorry.

PR #290, opened 2019-11-12
Fix references to PaletteMap, which was renamed to Attributes

jagger27:


Rangi42:
That variable is just a palette.

jagger27:
It's not... it refers to the variables declared at the beginning of the Attributes scope. Blame shows it was renamed a few months ago.

Rangi42:
You're right, thanks! It's fixed; I also changed the .blks to .ablks.

PR #294, opened 2019-11-23
fix game crash on opening bag

### LINKED ISSUE #293: Crash in non-faithful when opening bag
Introduced in commit https://github.com/Rangi42/polishedcrystal/commit/ca8718b338190cb6357a938cdc28040eca96ca23

Demonstration: https://www.youtube.com/watch?v=UGdJgdQoZ84

Reproduction: build above commit, run in any emulator, open bag at any point.

itsdarsh:
Fixes #293 - this wasn't specific to faithful, the issue is that all pointers to NoItemIcon for items past index $EA (including $FF) were removed, so arbitrary code was being executed upon decompressing that sprite. There was a few ways I could've done this but for now I decided to just have it where $FF is checked for in wCurItem, and if it's found then the item icon at pointer $00 is used, which is NoItemIcon. Note that any glitch items $EB to $FE will still crash the game upon opening their icon, so if that's undesirable then use the following code:

`    ld a, [wCurItem]`
`    cp MIRAGE_MAIL + 1`
`    jr c, .has_icon`
`    xor a`
`.has_icon`

in place of the code I used. Note that this will cease to work if an item has an index higher than Mirage Mail ($EA as of this writing). This issue does not occur with Key Items, which handles the cancel option differently.

I also fixed a minor bug where key item descriptions would not show up in the bag.

### DIFF
--- a/gfx/items.asm
+++ b/gfx/items.asm
@@ -19,7 +19,11 @@ UpdateItemIconAndDescriptionAndBagQuantity::
 	farcall UpdateItemDescriptionAndBagQuantity
 UpdateItemIcon::
 	ld hl, ItemIconPointers
-	ld a, [wCurSpecies]
+	ld a, [wCurItem]
+	inc a
+	jr z, .cancel
+	dec a
+.cancel
 	ld e, a
 	ld d, 0
 	add hl, de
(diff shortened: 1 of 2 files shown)


Rangi42:
Thanks! Can you please modify this to use `NUM_ITEMS + 1`, not `MIRAGE_MAIL + 1`?

PR #296, opened 2019-11-27
Add build script

jagger27:
Adds a build script that builds a "nightly" release on push to master. It automatically prunes the last release tagged as `nightly`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -56,7 +56,7 @@ jobs:
       with:
         upload_url: ${{ steps.create_release.outputs.upload_url }}  
         asset_path: ./polishedcrystal/build/polishedcrystal-3.0.0-beta.gbc
-        asset_name: polishedcrystal-3.0.0-${{ github.sha }}.gbc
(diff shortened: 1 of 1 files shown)


Rangi42:
`uses: jagger27/delete-release@master` should refer to `Rangi42`.

Rangi42:
Also, please append `$GITHUB_SHA` (the commit hash) to the ROM names so users can identify where it came from.

jagger27:
> `uses: jagger27/delete-release@master` should refer to `Rangi42`.

You would need to fork [jagger27/delete-release](https://github.com/jagger27/delete-release) if you want that. As it is, it's reusable action to delete a release. I added the SHA suffix. That environment value keeps the long hash, mind you, so it's a bit ugly.

Rangi42:
Thanks! `jagger27/delete-release@master` is fine then.

You're right about the long hashes; `${GITHUB_SHA::8}` evaluates to the first 8 characters.

jagger27:
Alright, should be good to go! Sorry for the commit spam. Gotta make a commit every time I test the build rule. Once it's in this repo the hash will actually match the commit.

PR #297, opened 2019-11-30
Continue sentences instead of new paragraphs

JanJastrow:
Hey, since polished crystal 3.0 is now in beta, I though to give it another run.
I stumbled upon these texts directly at the start of the game and I always thought I doesn't make sense to have new paragraphs instead of just continuing the sentence.
Please let me know what you think about this PR :)

### DIFF
--- a/data/text/common_3.asm
+++ b/data/text/common_3.asm
@@ -18,9 +18,8 @@ if !DEF(DEBUG)
 
 	para "A world of dreams"
 	line "and adventures"
-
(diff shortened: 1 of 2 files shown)


Rangi42:
One `para`, one `line`, and then `cont` until the end of the sentence was RBY's style; GSC seems to more consistently split a sentence into two "paragraphs" if it fits that way. I prefer that style, since it's fewer A presses. Thanks for the suggestion, though.

PR #300, opened 2019-12-01
fix crash on using FailText_CheckOpponentProtect

### LINKED ISSUE #263: Teleport while Transformed crashes
probe: "unknown opcode error game hangs when opponent ditto transform into abra and it uses teleport"

### LINKED ISSUE #266: Roar while opponent is using Dig crashes game
Specifically, when an NPC uses Roar while the user’s Pokémon is using Dig the game crashes.

itsdarsh:
Fixes #266 - game would (sometimes) crash when using Roar on a flying/digging mon because text wasn't properly loading text when wAttackMissed = 1. Also, the constant ATKFAIL_CUSTOM should actually have a value of zero, not six to match up with the jump table set up. I know there's a TODO to use it more often and I believe it is now correctly set up. I do think it should use a more generic name, but I can't immediately think of one, sorry.

### DIFF
--- a/constants/battle_constants.asm
+++ b/constants/battle_constants.asm
@@ -154,13 +154,13 @@ endc
 
 ; attack failure modes, higher gives priority
 ; TODO: complete this functionality
-	const_def 1
+	const_def
+	const ATKFAIL_CUSTOM  ; custom message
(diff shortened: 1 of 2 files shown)


FredrIQ:
A failure state of 0 shouldn't be a failure at all (i.e. a success).

PR #309, opened 2019-12-07
Updated Kurt's text to reflect instant ball creation

JanJastrow:
I think since Kurt will now create the balls instantly for you, his text should be updated.

I have also noticed that there more lines in ```KurtsHouse.asm```that do not make sense any longer, but that's something for another commit.


Rangi42:
Thanks, but it does take "a while" in-game; the player walks away from Kurt and it fades to black. The player gets the Ball instantly, but time is implied to pass.

PR #312, opened 2019-12-08
Tackled TODO: Give Lyra Custom Movesets

flempa316:
These party setups should make battles against Lyra more competitive, but will require some minor pokemon learnset adjustments... I can post those adjustments too if these changes look good!


Rangi42:
Thanks! Please do post the adjustments; I may change the learnsets, or change her movesets.

Rangi42:
So, you can just keep adding more commits to the `patch-1` branch and they'll get added to this pull request. When you make `patch-2`, `patch-3`, etc they're separate PRs, which is not correct since they're all related.

PR #361, opened 2020-01-02
Update parties.asm

Tenmu76:
Updated some parties for Battle Tower. Changes begin on line 40.


FredrIQ:
Looks mostly good, but Dugtrio needs an endc to finish the faithful checks.

PR #367, opened 2020-01-07
Shamouti Spark Battle Text

flempa316:
Took a stab at writing out the Spark text for each possible scenario the player would have when battling him.


Rangi42:
Thank you, but there are no beacons; the birds and Lugia are already available from the Lawrence event. The three PGo team leaders are just going to be optional side battles.

flempa316:
Ahh okay!

PR #383, opened 2020-02-04
Added the ability to change a pokémon's ball

wolfboyft:
You can now replace the pokeball a 'mon is in
Though in the code it's called "caught ball" implying that it's just catching records, so maybe i shld change that?
also changed the sprites but i doubt that;d be a welcome change so i havent commited that
(removed the L shaped corners. slightly more "octogonal" now but preferable IMO)
You don't get the ball back bc you could farm master balls with breeding etc and also, like, "you don't get paid back for your previous haircut"


data/text/common_3.asm
```diff
@@ -1103,3 +1103,8 @@ ItemCantBeSelectedText::
 	text "That item can't be"
 	line "selected."
 	prompt
+
+AlreadyInThatBallTextData::
+	text "Your #mon is in"
+	line "that ball already."
```
Rangi42: Capital "Ball".

engine/item_effects.asm
```diff
@@ -314,6 +314,9 @@ KeyItemEffects:
 	dw ApricornBox      ; APRICORN_BOX
 
 PokeBallEffect: ; e8a2
+	ld a, [wBattleMode]
+	cp 0 ; overworld
```
Rangi42: `and a` is the standard [optimized](https://github.com/pret/pokecrystal/wiki/Optimizing-assembly-code#compare-a-to-0) way to test for 0 instead of `cp 0`.

engine/item_effects.asm
```diff
@@ -2313,6 +2316,37 @@ ItemWasntUsedMessage:
...
+	ld a, [wCurItem]
+	ld b, a
+	ld a, [wCurPartyMon]
+	ld hl, wPartyMon1CaughtBall
+	call GetPartyLocation
+	ld a, [hl]
+	and CAUGHTBALL_MASK
+	cp b
+	
+	jp z, AlreadyInThatBallMessage
```
Rangi42: `jr` is a more [optimized](https://github.com/pret/pokecrystal/wiki/Optimizing-assembly-code#relative-jumps) way to make short jumps. `AlreadyInThatBallMessage` is close enough to use `jr` instead of `jp`. Also, style-wise I'd expect the blank line to be after the jump, not before, since it separates the next "paragraph" of code which handles the swap.

engine/item_effects.asm
```diff
@@ -2313,6 +2316,37 @@ ItemWasntUsedMessage:
...
+	ld a, [hl]
+	and -CAUGHTBALL_MASK-1 ; One's complement
```
Rangi42: The typical way to write this is `and $ff ^ CAUGHTBALL_MASK` (or sometimes `and $ff - CAUGHTBALL_MASK` from before rgbasm handled byte xor correctly).

engine/item_effects.asm
```diff
@@ -2313,6 +2316,37 @@ ItemWasntUsedMessage:
...
+	add b
+	ld [hl], a
+	call UseDisposableItem
+	ld hl, BallReplacedText
+	jp PrintText
+
+BallReplacedText:
+	text "Ball changed."
```
Rangi42: It would be nice if this were more descriptive. Like "Transferred <mon name> to the <item name>!" Or at least "Stored in the <item name>!"

Rangi42:
Thanks for submitting this! It's a neat feature that I'm definitely considering to keep. However, I like the graphics as they are; IMO they look more rounded that way.

If I were going to switch to these graphics, there's one more file that would need updating for consistency: gfx/battle_anims/pokeball.2bpp.lz. (`python gfx.py png gfx/battle_anims/pokeball.2bpp.lz` recreates the PNG to make it editable.)

Rangi42:
You may as well edit CREDITS.md too to add yourself to the Code section. ;)

PR #386, opened 2020-02-06
Trainer view occlusion fix thing

wolfboyft:
Doesn't actually work yet btw
I wonder whats wrong with it?
i had to free up some space in home.asm so that i could jump to a routine elsewhere btw


wolfboyft:
I accidentally did this completely wrong. sorry!

PR #387, opened 2020-02-07
Strength issue fix: Not fully tested yet, but should work.

wolfboyft:


wolfboyft:
We'll need to make a map to test all the cases. In the meantime, how is the code quality?

engine/overworld/map_objects.asm
```diff
@@ -3061,3 +3061,134 @@ PRIORITY_HIGH EQU $30
...
+CheckOccludingObjects:: ; here for space reasons
+	push de
+; returns carry = occluded
+
+	; Save location of current trainer
+	ld a, [wCurTrainer]
+	ld bc, OBJECT_LENGTH
+	ld hl, wMapObjects + MAPOBJECT_X_COORD
+	rst AddNTimes
+	ld a, [hl]
```
Rangi42: Instead of `ld a, [hl]` and then `dec hl`, this can just be `ld a, [hld]`.

engine/overworld/map_objects.asm
```diff
@@ -3061,3 +3061,134 @@ PRIORITY_HIGH EQU $30
...
+
+.checkYCoords
+	ld a, [wCurTrainerY]
+	ld b, a
+	ld a, [wPlayerStandingMapY]
+	ld c, a
+	ld a, [wCurPotentialOccluderY]
+
+.check
+	ld b, b
```
Rangi42: No need for `ld b, b`.

home.asm
```diff
@@ -964,7 +965,11 @@ CheckTrainerBattle:: ; 360d
 	pop de
 	pop bc
 	and a
-	jr z, .startbattle
+	jr nz, .next
+	
+; And is unoccluded
+	farcall CheckOccludingObjects ; rst farcall says it retains "a, bc, de, and hl" but return farcall says it tries to save f?
```
Rangi42: That comment is referring to the register values of the caller. Commit 90a7bfbbee7303b681cdcaaca0ccf402013d7575 stopped preserving the caller's flags, but still preserves the flags when returning.

home/init.asm
```diff
@@ -94,13 +94,11 @@ Init:: ; 17d
 
 ; Initialize the RNG state. It can be initialized to anything but zero; this is just a simple way of doing it.
 	ld hl, wRNGState
-	ld a, "R"
+	ld a, 1
```
Rangi42: Is this necessary? IMO just leave the RNG state as-is, it's only a few bytes more, and I'd rather not change its behavior.

wolfboyft:
Sorry abt tbe breakpoint.
Optimisations noted btw

Also the RNG change was to make space for the call and jump. if an optimisation can be msde elsewhwrw rhatd be good ofc. i agree tjat changing tje rngs bbehaviour is ugly

Rangi42:
If it's needed for space and there's no better reduction, then go ahead and leave the RNG change. I'll keep an eye out for bank 0 space.

Rangi42:
I've removed `IsInJohto` from home, so no need to change the RNG.

wolfboyft:
Oh! I also changed ClearsSRAM (ClearSRAM), ClearVRAM and ClearWRAM to inlines and not calls (as macros) because they're only used once. But that's not necessary now. However, if more home space is required, that could work. It's a good way to deal with single-use functions without ruining the source code.

wolfboyft:
Now, I'll do your changes, and make a test map.

home/init.asm
```diff
@@ -85,19 +130,22 @@ Init:: ; 17d
 	pop af
 	ld [hCGB], a
 
-	call ClearWRAM
+	ClearWRAM
```
Rangi42: One-off macros for whole chunks of code is not a style used in the pret disasm projects. If you want to inline these, literally just inline them; replace that `ClearWRAM` macro invocation with the actual code, with blank lines setting it apart as its own "paragraph", and maybe a comment on top like `; clear WRAM`. Likewise for the other two.

engine/overworld/map_objects.asm
```diff
@@ -3061,3 +3061,132 @@ PRIORITY_HIGH EQU $30
 	dw wObject11Struct
 	dw wObject12Struct
 ; 5ae8
+
+CheckOccludingObjects:: ; here for space reasons
```
Rangi42: This could go in its own engine/overworld/check_occluding_objects.asm file, and then get INCLUDEd in main.asm underneath this file's INCLUDE.

Rangi42:
(This was not actually working when I tested it; it was preventing trainers from seeing you at all in some cases, even when standing right in front of them.)

PR #462, opened 2020-04-04
Sing is effecting ghost types

flempa316:
Sing Effects Ghost Types
This seems to fix the issue


flempa316:
Sing Fix
Here it is after implementing the change

SadisticMystic:
This is not a bug. Sing, along with every status move in the game except Thunder Wave, does not check the type chart, and this is official behavior.

Issue #455, opened 2020-03-31
Struggle can be Disabled


Rangi42:
Dare Evans: "Disabling a Pokemon with one move disables Struggle. Disabled Magikarp's Splash and then it said Struggle was disabled. Build is 9d2d96ec."

PR #463, opened 2020-04-04
fix #454 (Minimize) and #455 (Disable)

### LINKED ISSUE #454: Entering and exiting the Bag during battle resets any Minimized sprites
image

itsdarsh:
Issue #455 is related to how Disable is checked in PC: the first nibble of `wPlayerDisableCount` is the move that is disabled, the second nibble is the number of turns left. When a Pokemon only has one move, the first nibble will always end up being $1 since only the first move exists (stored in memory as move $0 in address `wCurPlayerMoveNum`, the first nibble is decremented before being compared to the used move). The Pokemon is then forced to select move $0 again since its other moves are all empty, and the game properly catches that the move is disabled and thus sets the used move to Struggle. However, when the game then compares `wCurPlayerMoveNum` to the decremented first nibble of `wPlayerDisableCount`, these values still match up, so the game is tricked into saying that Struggle must be disabled. The fix is to set `wCurPlayerMoveNum` to -1 ($FF) so that check always fails and Struggle goes on as normal. In my testing, in the case of Struggle a move number of -1 does not have an issue. This is true for `wEnemyDisableCount`/`wCurEnemyMoveNum` as well.

Issue #454 is a bit deeper than a graphical glitch, aside from lowering evasion none of the other intended features of Minimize actually work properly. This is because `wPlayerMinimized`/`wEnemyMinimized` are never actually set, so anything that checks them automatically fail. This includes properly redrawing the correct battle sprite and Stomp/Body Slam not getting 2x power or perfect accuracy against a minimized target. The least hacky way to set this IMO is to have the minimized substatus be part of `BATTLE_VARS_SUBSTATUS_2`, and have any caller that needs this value get the battle var instead of choosing between `wPlayerMinimized` and `wEnemyMinimized`. Being minimized is a boolean value, so separate wram addresses are a bit excessive. If it's desired to instead have more free space in BATTLE_VARS_SUBSTATUS_2 then that's easy to revert, so it's up to you.

Also made a couple small bug fixes and optimizations as well. Notably, even if `wPlayerMinimized` was properly set, the perfect accuracy buff for Stomp and Body Slam do not work as intended - in fact, Stomp and Body Slam will have perfect accuracy if the _user_ is minimized, not the opponent. Also, the AI will prioritize Stomp against a minimized opponent, but not Body Slam. Finally, since `EFFECT_MINIMIZE` is now separate from `EFFECT_EVASION_UP_2` I added it individually to the AI setup scoring for the time being, and I think `AI_Setup` deserves a bit of a look of its own.


Rangi42:
Thank you! Using `BATTLE_VARS_SUBSTATUS_2` is fine.

PR #464, opened 2020-04-05
fix HP bar inconsistency issues and Rampage/Own Tempo bug

### LINKED ISSUE #445: HP bars are inconsistent
Gabe: "I KOed Will's Xatu then the HP bar went back to 1"

SadisticMystic:

> vanilla notably attempts to redraw the health bar after every individual HP lost (with the side effect that dealing huge amounts of damage to something like Red's Snorlax takes an unusually long time), and the rounding algorithm at every point is the same--drawing HP bars can never round down to 0 pixels, even when you have 1/346
>
> PC takes out the throttler to allow health bars to scroll down in reasonable time even when HP gets large, but uses a different algorithm for rounding while it's mid-subtraction (this one can round down to 0 pixels) compared to the one it uses at the end when the health bar needs to be fixed in place going forward

Possibly related to #434.

itsdarsh:
Fixes #445 

There's already a pretty good description of what's happening thanks to @SadisticMystic, but there's a further bit of nuance. `ComputeHPBarPixels` actually does have a failsafe where if a HP bar would have 0px filled but the Pokemon isn't fainted, then 1 is returned instead of 0. This normally works fine, but because `_AnimateHPBar` actually needs twice that value due to 60fps updating of numbers converting to 30fps bar scrolling, `ComputeHPBarPixels` essentially returns 1/2 instead, which rounds down to 0. Thus, until the HP bar is later redrawn, it temporarily shows as 0px. This fix just adds a bit of an extra failsafe to ensure that 1 is incremented to 2 so the endpoint is properly set.

There was also a minor bug where the opponent's Own Tempo was incorrectly checked for confusion prevention over the user's, so that's addressed here as well.


engine/battle/effect_commands.asm
```diff
@@ -1902,6 +1902,7 @@ BattleCommand_checkhit:
 	ret
 
 .NoGuardCheck:
+;Returns z if either the user or opponent has the No Guard ability
```
Rangi42, line 1905: Nitpicking: Comments always have a space after the `;`. Same in `.AntiMinimze`.

itsdarsh: oops :V

Issue #457, opened 2020-03-31
PC Box data in saves is somehow sometimes corrupted


Rangi42:
This may be related to the battle with Leaf, and/or to `sScratch`. It's not clear when or how this occurs.

PR #474, opened 2020-04-10
a few bug fixes, including a save "corruption" and a crash

### LINKED ISSUE #471: Escape rope crashes the game
Pretty self-explanatory.

When used in dark cave :
image
When used in sprout tower : 
image

itsdarsh:
**Addresses (maybe part of?) #457** - In `InitDisplayForLeafCredit` the game attempts to fill `wBGPals1` and `wOBPals1` with white palettes. However, there is never a wram bank switch to the correct bank first, so instead of overwriting those addresses in bank 5 it overwrites those addresses in bank 1 instead. This causes it to overwrite the first byte of `wPlayerID`, and when the player attempts to save the game will assume the player is playing on a new save file due to this new TID and will wipe the old save info if the player picks yes. _Note that this bug also exists in vanilla Crystal for Red's endgame credits._ However, since `wBGPals1` and `wOBPals1` have different physical locations in vanilla, they're just overwriting some temporary buffers. All that said, since the code did nothing in vanilla it does nothing here, so I decided to just remove it.

**Fixes #471** - Due to an n-gram, "Escape Rope@" now takes up 11 bytes in memory (and for the purposes of this explanation, in `wStringBuffer1`), not 12. That happens to be the same max length as a Pokemon nickname, so when `EscapeRopeOrDig` calls `GetPartyNick` for an invalid `wCurPartyMon` (in the case of escape rope) it is essentially writing 11 bytes of $00 into `wStringBuffer1`, overwriting the terminating character left from "Escape Rope@". Thus when the game attempts to copy this "nickname" to other string buffers, unless the dungeon name is longer than 11 bytes in memory (i.e. Slowpoke Well). This is _not_ a bug in vanilla because while it still pulls up an invalid nickname, it has a sanity check (`CorrectNickErrors`) to replace a nickname of all $00 with "?@?????????". Instead of reincorporating that function I just moved the check for if an escape rope is being used to be before the call to `GetPartyNick` since that's what should've been done in the first place :V

**Fixes attract "succeeding" after target is already attracted** - `SUBSTATUS_IN_LOVE` was checked by the battle engine, but nothing was done with that information

**Fixes "Seen" string in Pokedex page being displayed improperly** - `Pokedex_PlaceString` does not factor in n-grams, but "Seen" was stored as a `db` so "en" would've been stored as an n-gram. Thus, it was parsed literally as char $27 (which is part of frontpic data in this case), leading to the inaccurate display.

**Fixes hatchling frontpic showing the wrong mon** - not much more to say other than a `ld` was backwards

**Fixes possible issue with `BillsPC_InitRAM` going out of bounds** - instead of a hardcoded loop counter of $338, a more dynamic counter of `wBillsPCDataEnd - wBillsPCData` is used. Unclear if this ever caused any problems in the first place

### DIFF
--- a/contents/bank_ends.txt
+++ b/contents/bank_ends.txt
@@ -5,7 +5,7 @@
 > -- Morimoto, Pokémon Ultra Sun/Ultra Moon
 > <https://www.serebii.net/ultrasunultramoon/virtualconsole.shtml>
 
-Free space: 280255/2097152 (13.36%)
+Free space: 280264/2097152 (13.36%)
 
 bank	end	free
 $00	$3dfd	$0203
@@ -107,7 +107,7 @@ $5f	$8000	$0000
 $60	$7fff	$0001
 $61	$8000	$0000
 $62	$7fff	$0001
-$63	$8000	$0000
+$63	$7fff	$0001
 $64	$8000	$0000
 $65	$8000	$0000
 $66	$8000	$0000
@@ -118,7 +118,7 @@ $6a	$8000	$0000
 $6b	$7fff	$0001
 $6c	$8000	$0000
 $6d	$8000	$0000
-$6e	$7b54	$04ac
+$6e	$7b4c	$04b4
 $6f	$4000	$4000
 $70	$4000	$4000
 $71	$4000	$4000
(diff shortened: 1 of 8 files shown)


Rangi42:
Those first two bugs are great. I'll start a doc or wiki page for pokecrystal for "not-quite-bugs".

PR #488, opened 2020-04-15
Fixes issue 480

flempa316:
Dazzlingleam and rock blast fix


Rangi42:
There are probably more animations besides those three with this sort of problem, but this is a good start.

PR #494, opened 2020-04-19
refactor hazards/screens/guards and add modern mist mechanics

### LINKED ISSUE #433: Upgrade Guard Stats/Guard Spec (Mist) to Gen VII mechanics
It's now a screen-like effect, not a volatile status condition.

itsdarsh:
**Fixes #433 -** Mist now works as a 5-turn static effect, similar to Safeguard. Currently it's only applicable to Guard Spec but can work with a re-implementation of the move Mist as well, should it ever come to that.

**Uses guards/screens counters as flag for if they're online -** basically, if they're non-zero, the effect is in action. Didn't really make a ton of sense for them to also have a separate flag to mark if they're active, so `wPlayerHazards` and `wEnemyHazards` can now contain flags for other hazards like Stealth Rock and Sticky Web, should they be implemented.


engine/battle/endturn.asm
```diff
@@ -827,105 +828,105 @@ HandleLeppaBerry:
...
+	call SwitchTurn
 
-.enemy_first
-	call .CheckEnemy
-.CheckPlayer:
-	call SetPlayerTurn
-	ld de, .Your
-	call .Copy
+.do_it
+	ld a, [hBattleTurn]
```
Rangi42, line 837: `ldh a, [hBattleTurn]`

engine/battle/endturn.asm
```diff
@@ -827,105 +828,105 @@ HandleLeppaBerry:
...
+	call SetFastestTurn
+	call .do_it
+	call SwitchTurn
 
-	call .CheckEnemy
-.CheckPlayer:
-	ld a, [wPlayerScreens]
-	bit SCREENS_SAFEGUARD, a
+.do_it
+	ld a, [hBattleTurn]
```
Rangi42, line 852: `ldh a, [hBattleTurn]`

engine/battle/move_effects/brick_break.asm
```diff
@@ -2,28 +2,24 @@ BattleCommand_brickbreak:
...
 	call StdBattleTextBox
-	pop bc
 	pop hl
-.light_screen_done
-	inc bc
-	bit SCREENS_REFLECT, [hl]
+.reflect_done
+	pop af
+	and SCREENS_LIGHT_SCREEN
+	ld a, 0 ; zero flag needs to be preserved
```
Rangi42, line 21: Does this need to alter `a`, or could it do `ld [hl], 0` instead of `ld a, 0 / ld [hl], a`?

itsdarsh: no it doesnt need to alter `a` at all here but i was doing my best game freak impersonation and overengineering it :V

PR #496, opened 2020-04-21
update rampage to modern mechanics

### LINKED ISSUE #415: Thrash/Outrage lock in even when the first use fails
SadisticMystic: "Pokemon are becoming locked into Thrash or Outrage even if the first turn goes into a type immunity (that's not supposed to create any rampage and confusion never happens because of it)"

itsdarsh:
**Fix #415 and update rampage mechanics -** rampage moves (e.g. Outrage) will now no longer be locked into a rampage move if the attack misses for any reason or if the user otherwise cannot move. Note that the user will still get confused on the last turn of the rampage even if their attack otherwise missed. Plus, the game will now tell the player if the user is confused out of fatigue (this did not happen in vanilla Crystal)

**dereference `wSomeoneIsRampaging` -** I'm not even sure this did anything productive in vanilla but it definitely does nothing here.

**fix potential issues with `CantMove` -** `CantMove` tries to check for `SUBSTATUS_ROLLOUT` in the user's 3rd substatus byte, but the flag for rollout is actually stored in the first byte. I don't really know what kind of issue this would've caused in general, but it probably would've affected Rollout and being unable to move and maybe also Fly and Substitute. Whatever the problem was, it won't show up from here.

**lock-in moves now end if there's not target and update turns counters -** This actually was an issue before.

**merged related substatuses into one variable -** Basically, substatus 3 contains any variable that would relate to taking multiple turns, and also confusion as that's related to rampaging.


itsdarsh:
If I don't get to it first tomorrow morning remind me to rewrite some of the confusion stuff because Safeguard should not prevent confusion due to fatigue and at that point the only check that's really necessary is for Own Tempo and if the user is already confused, both of which are probably handled better in `HandleRampage` anyway (there are no items that prevent confusion in vanilla afaik)

itsdarsh:
`CantMove` wasn't actually clearing bits properly since they were formatted incorrectly (SUBSTATUS_ constants contain an integer from 0-7 and do not represent a bit) so it should actually work better now. Otherwise I just made confusing the user in `HandleRampage` more efficient and intuitive, though it loses the ability to preserve `hl` which only matters in one case anyway.

PR #498, opened 2020-04-24
relabel battle anim constants/animation branches

itsdarsh:
Matched battle anim constants and animation branches to pokecrystal.

**NOTE:** I also gave labels to a lot of battle obj and battle function constants that pokecrystal left as a generic hex value. I didn't do all of them since I wasn't too certain what to call them, but please let me know if I should revert those renamed labels back to what pokecrystal had, or if I should give them different names instead. I did _not_ change any constant names that pokecrystal already had, just made new ones.


constants/battle_anim_constants.asm
```diff
@@ -24,196 +25,201 @@
 	const BATTLEANIMSTRUCT_16
 	const BATTLEANIMSTRUCT_17
 BATTLEANIMSTRUCT_LENGTH EQU const_value
+NUM_ANIM_OBJECTS EQU 10 ; see wActiveAnimObjects
```
Rangi42, line 28: If the code doesn't use these constants, then don't copy them.

constants/battle_anim_constants.asm
```diff
@@ -24,196 +25,201 @@
...
 
+; Start tile for battle animation graphics
+BATTLEANIM_BASE_TILE EQU 7 * 7  ; Maximum size of a pokemon picture
```
Rangi42, line 31: ditto

itsdarsh:
Those labels do still have uses in polished code, so I just ported that over from pokecrystal as well. There's also a bug from vanilla that can be fixed with those constants, but to preserve md5 hash I'll just leave it alone until this pr is closed.

Rangi42:
Thanks! May as well fix it in this one, the hash gets changed all the time anyway.

itsdarsh:
Fixed.

PR #547, opened 2020-07-18
Fix some Spanish dialogue

aaaaaa123456789:
This fixes some spelling mistakes in that particular trainer's text. Let me know if there are any others and I'll check them as well.


Rangi42:
Thank you!

PR #573, opened 2020-09-16
replace wEngineBuffers with useful variable names

itsdarsh:
Also replaced `ENVIRONMENT_5` with `ISOLATED`.

NOTE: There are two bugs I uncovered in the process of the refactor. That said, instead of fixing these bugs, I decided to leave them as is for two reasons: one, to ensure the md5 has was the same as upstream; two, because I'm gonna be rewriting the relevant segments anyway when converting the phone contact list to a bit array. It's up to you, but for those two reasons I'd prefer to just merge the changes without actually addressing them for the time being.


data/maps/environment_colors.asm
```diff
@@ -4,7 +4,7 @@ EnvironmentColorsPointers:
 	dw .OutdoorColors ; ROUTE
 	dw .IndoorColors ; INDOOR
 	dw .DungeonColors ; CAVE
-	dw .Perm5Colors ; ENVIRONMENT_5
+	dw .Perm5Colors ; ISOLATED
```
Rangi42, line 7: `dw .IsolatedColors`

Rangi42 (changes requested):
Just needs to rename `.Perm5Colors` to `.IsolatedColors`

PR #585, opened 2020-11-27
Allowing Toolchain to work properly on OSX

Cl0v1s:
MacOSX doesnt come with md5sum (it has a pre-installed md5 utility but it hasnt -c option required for `make compare`)

Since md5sum is required to `make` and `make compare`, this PR add instructions to install it on OSX.

Moreover the script tools/collision_asm2bin.sh uses `tail` and `head` with negative -c param, which is not supported by embeded head and tail binaries, causing buggy/missing collisions ig.
`head: illegal byte count -- -8` are visible in `make` logs.

ref:
https://unix.stackexchange.com/questions/169079/negative-arguments-to-head-tail
https://www.garron.me/en/bits/how-to-md5sum-mac-os-x.html
https://raamdev.com/2008/howto-install-md5sum-sha1sum-on-mac-os-x/

### DIFF
--- a/tools/collision_asm2bin.sh
+++ b/tools/collision_asm2bin.sh
@@ -3,15 +3,15 @@
 
 TEMP_ASM=$(mktemp collision.asm.XXX)
 TEMP_O=$(mktemp collision.o.XXX)
-OS=$(uname -s)
 
 HEAD=head
-tail=tail
+TAIL=tail
 
-if [ $OS = "Darwin" ]
-then
-    HEAD=ghead
-    TAIL=gtail
+if ! type ghead > /dev/null; then
+  HEAD=ghead
+fi
+if ! type gtail > /dev/null; then
+  TAIL=gtail
 fi
 
 echo 'INCLUDE "constants/collision_constants.asm"' > "$TEMP_ASM"
--- a/tools/collision_asm2bin.sh
+++ b/tools/collision_asm2bin.sh
@@ -7,10 +7,10 @@ TEMP_O=$(mktemp collision.o.XXX)
 HEAD=head
 TAIL=tail
 
-if ! type ghead > /dev/null; then
+if type ghead > /dev/null; then
   HEAD=ghead
 fi
-if ! type gtail > /dev/null; then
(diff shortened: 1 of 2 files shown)


FredrIQ:
md5sum is only required for some of the Makefile commands (such as freespace).

I do not recognize ghead/gtail. Is it some OSX thing?

Cl0v1s:
md5sum is used at the end of the basic `make` call without any option, without it, the make crashes :/

All gnu utils are install by homebrew (apt-get of OSX, sort of) with the g-prefix to avoid messing with "classical" OSX installation. 
gtail and ghead are exactly the same binaries that the ones used in cygwin :)

tools/collision_asm2bin.sh
```diff
@@ -3,12 +3,22 @@
 
 TEMP_ASM=$(mktemp collision.asm.XXX)
 TEMP_O=$(mktemp collision.o.XXX)
+OS=$(uname -s)
+
+HEAD=head
+tail=tail
```
aaaaaa123456789, line 9: This should be `TAIL=tail`.

tools/collision_asm2bin.sh
```diff
@@ -3,12 +3,22 @@
...
+
+if [ $OS = "Darwin" ]
```
aaaaaa123456789, line 11: Perhaps, instead of detecting the OS, detect if `gtail` and `ghead` exist? If those exist, it's very likely that you'd want to use them instead of `head`/`tail`, regardless of OS.
(This is basically a case of the general principle of "ask the question you need".)

aaaaaa123456789:
If `md5sum` is used unconditionally, just replace it with a command that always exits successfully (such as `true`) when it's not available.

Cl0v1s:
Is it desirable to bypass md5 hash generation ?

Rangi42:
Since macOS requires `coreutils` anyway to get `ghead` and `gtail`, and that comes with `md5sum` too, then I think the current setup of using `md5sum` is fine. The method of conditionally using `ghead` and `gtail` also looks good. Thanks!

PR #611, opened 2021-03-23
Battle factory

FredrIQ:
Implements the Battle Factory. These are the mechanics:

* Each challenge run features 7 trainers (similar to the Battle Tower).
* When you enter a challenge, you get 6 randomly generated Pokémon with the difficulty level being equal to what the 6th trainer in that challenge run would use. The difficulty level of your rental Pokémon is elevated by (total swaps performed the current streak) / 6, capped at 8. This means that the more swaps you do over the course of a streak, the stronger Pokémon you are given as initial 6. Opposing trainers do not get a bonus like this.
* After beating a Trainer, you may choose to swap one of your Pokémon with one of the ones used by the last foe you met.
* Before you select your initial team, or choose to swap one of your current mon with one from the last foe, the battle judge does "research" on the upcoming Trainer. The research info is as follows:
* * On the first challenge run, you get the list of all 3 Pokémon they use
* * On the second run, you get the list of the first 2 Pokémon they use
* * On the third run, you are told their lead, and one of their moves (the first one in the movelist).
* * On the fourth run, you are not told the lead, but you are still told the first move that the lead has.
* * On the fifth and later runs, you are told what Type they favour the most.
* Due to the nature of rental luck, earlier runs run the AI at reduced strength. The AI levels go as follows:
* * Challenge run 1: Master type matchups.
* * Challenge run 2: Avoid status moves into immunities.
* * Challenge run 3-4: Use the strongest damaging move.
* * Challenge run 5+: Smart AI -- the AI runs at full strength from this point onwards.
* Every 3rd (successful) challenge run sees you battle the Factory Head.
* A successful run awards a Mint Leaf.


Rangi42 (changes requested):
Amazing work, thank you! Just a few constant changes and then you can merge any time:

- `BATTLETOWER_TYCOON` → `BATTLETOWER_TOWERTYCOON`
- `BTCHALLENGE_TYCOON` → `BTCHALLENGE_FACILITYBRAIN`
- `50` → `BATTLETOWER_FORCED_LEVEL` (in `engine/events/battle_tower/load_trainer.asm` and maybe other files)

Rangi42 (approved):

PR #626, opened 2021-04-17
Implemented SoupPotato's Pokemon sprites + anims

teddybuns:
Implemented all of SoupPotato's Pokemon sprites + anims from their sprite codebase, taking into account Pokemon that have been (temporarily) removed from PC and those with formes and variants that needed to have their folders renamed/adjusted accordingly. Excluded Arbok and Magikarp due to having multiple different forms.


Rangi42:
Superseded by #733.

PR #653, opened 2021-10-31
New BT sets for 9bit update

Salamandar73:


FredrIQ:
Merged manually to avoid the unnecessary added files. Thanks!

PR #663, opened 2022-01-16
First implementation of Lyra telling you about Pokémon evolution methods

rik-smeets:
This PR is inspired by issue #652 (Make calling Lyra useful). If the player calls Lyra when Magneton is in the party, she wil tell you about it's evolution method.

I would like to further extend this for other Pokémon, but thought it was a good idea to first ask for any feedback here.

Issue #652, opened 2021-10-28
Make calling Lyra useful


Rangi42:
Call Lyra to learn about how Pokémon evolve. (She could say "I learned from Prof. Elm..." or "My grandma/pa told me..." since she's the daycare couple's granddaughter and Prof. Elm's assistant.) (Or just call some professor directly?)

https://bulbapedia.bulbagarden.net/wiki/Professor_Juniper#Evolution_hints_.28Xtransceiver.29

All B2/W2 evolution advice: https://pastebin.com/JnbKLmfu

PR #664, opened 2022-01-19
Add more Lyra evolutions tips

rik-smeets:
Builds upon #663 to work towards resolving #652. See the commits (or commit messages) for everything that has been added.

Lyra is now able to tell you about multiple evolution methods if multiple Pokémon in your party have certain evolution methods. She will ask beforehand if you'd like to hear about it. 

Also, I added a night time greeting for Lyra based on a bug report on Discord.

I created a separate file for Lyra's phone scripts, since it's already growing pretty fast and makes it easier to keep track of things. 

I'll keep adding more evolution methods, but prefer to keep pull requests small - especially if there are some changes in design/file organization, so I can get feedback by you. Please let me know if you prefer things different. Thanks!


Rangi42:
This is getting large enough that I think a more abstract approach would be appropriate.

One possibility would be for Lyra's script to check the species of your first party Pokémon, read its evolution data directly, and have a table of possible responses for each method.

Another would be to have a table of pairs of (species, pointer), where `pointer` goes to a custom script for each `species`. Any species not listed would have generic "it evolves at a high enough level" text. This might be more appropriate for a few reasons:

- It wouldn't require complicated code to parse the evolution method data
- More than half of all evolutions are by level, so could be ignored
- If the name of the Pokémon is read into `wStringBuffer1` with `GetPokemonName`, then the same pointed script could be used for all Water Stone evolutions, all `EVOLVE_HAPPINESS, TR_ANYTIME` evolutions, etc.
- Pokémon with more unusual evolutions, like unique or multiple methods, could have individual script text, instead of plugging their names and items/locations/parameters into generic text blocks.

I'm not asking you to take on such a project! Just saying that I don't plan to continue adding this kind of script-only logic for more and more species.

rik-smeets:
Yes, my solution only works for a limited number of Pokémon without becoming a huge chunk of code. Your approach sounds great, but unfortunately, my skills in assembly probably (well, definitely) aren't going to cut it.

Would you like me to abandon this PR? Or are you considering completing this for the time being, so Lyra can at least tell you about some of the more unusual evolutions?

Rangi42:
Just leave it open for now, you've written some good dialog that I could use.

Rangi42:
This feature has been implemented in 9bit.

PR #666, opened 2022-01-22
Fix flashing 'used move text' when using Fly/Dig

rik-smeets:
When using the move Fly or Dig, the text: `[Pokémon] used [Fly/Dig]!` flashes on screen for a split second before the animation begins. When using these moves, this text should not be shown at all (and it isn't in vanilla Crystal). For these moves, the text is shown after the animation. For example, `[Pokémon] flew up high!`.

I resolved this issue by changing the move's effect order.

### DIFF
--- a/data/moves/effects.asm
+++ b/data/moves/effects.asm
@@ -2149,8 +2149,8 @@ Teleport:
 Fly:
 Dig:
 	checkobedience
-	usedmovetext
 	doturn
+	usedmovetext
 	hastarget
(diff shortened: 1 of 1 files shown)


rik-smeets:
I just learned that this text ís shown in newer gens, on which the battle engine is based. Therefore closing this PR.

With instant text on, it is practically unreadable though, but that's probably a bit harder to fix and it's no biggie.

PR #668, opened 2022-01-23
Fix two bugs in battle animations

### LINKED ISSUE #665: Dig not bringing graphic back up after confusion self hit
As I recall vanilla behavior brought the sprite back up if something prevented the second part of dig from happening. That didn't happen and it led to not bringing it back up for the next attack.
It did bring it back up after the next round of digging.
The build I played with is based on stuff from back in October, and it looks like no one reported a similar issue.

https://www.twitch.tv/videos/1270334163

rik-smeets:
1. The animation that is shown when a Pokémon hits itself in confusion did not show. It seems this was accidentally removed in [this commit](https://github.com/Rangi42/polishedcrystal/commit/0c4a1e04e6900f9fdeac9f695fcb229d5e1ec45c#diff-88dbff1eb3920f69b17a3d5f613643e1a41096514706bc09f449e4aa637f20e4).
2. If the second turn of Dig/Fly fails (for example, when the Pokémon hits itself in confusion) the Pokémon's sprite stayed invisible. This was reported in #665. 

Resolves #665.

### DIFF
--- a/engine/battle/effect_commands.asm
+++ b/engine/battle/effect_commands.asm
@@ -456,7 +456,8 @@ EndTurn:
 OpponentCantMove:
 	call CallOpponentTurn
 CantMove:
-	call CheckRampageStatusAndGetRolloutCount ; ; hl becomes pointer to user substatus3
+	call .cancel_fly_dig
+	call CheckRampageStatusAndGetRolloutCount ; hl becomes pointer to user substatus3
 	jr z, .rampage_done
 	ld a, [de]
 	dec a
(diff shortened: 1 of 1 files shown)


engine/battle/effect_commands.asm
```diff
@@ -465,15 +466,19 @@ CantMove:
...
 	ld a, ~(1 << SUBSTATUS_RAMPAGE | 1 << SUBSTATUS_CHARGED | 1 << SUBSTATUS_FLYING | 1 << SUBSTATUS_UNDERGROUND | 1 << SUBSTATUS_ROLLOUT)
 	and [hl]
 	ld [hl], a
 	ret
 
+.cancel_fly_dig
+	ld a, BATTLE_VARS_MOVE_ANIM
+	call GetBattleVarAddr
+	cp FLY
+	jp AppearUserRaiseSub
```
Rangi42, line 478:
```suggestion
	jmp z, AppearUserRaiseSub
```

engine/battle/effect_commands.asm
```diff
@@ -465,15 +466,19 @@ CantMove:
...
+	cp DIG
+	jp AppearUserRaiseSub
```
Rangi42, line 480:
```suggestion
	jmp z, AppearUserRaiseSub
```

engine/battle/effect_commands.asm
```diff
@@ -465,15 +466,19 @@ CantMove:
...
+
```
Rangi42, line 481:
```suggestion
	ret

```

engine/battle/effect_commands.asm
```diff
@@ -465,15 +466,20 @@ CantMove:
...
 	ret
 
+.cancel_fly_dig
+	ld a, BATTLE_VARS_MOVE_ANIM
+	call GetBattleVarAddr
+	cp FLY
+	jmp z, AppearUserRaiseSub
+	cp DIG
+	jmp z, AppearUserRaiseSub
+	ret
```
Rangi42, line 481:
```suggestion
	cp FLY
	jr z, .fly_dig
	cp DIG
	ret nz
.fly_dig
	jmp AppearUserRaiseSub
```

PR #675, opened 2022-01-29
Add hidden nugget to Cherrygrove City

rik-smeets:
Adds the hidden nugget from HG/SS to Cherrygrove City. Includes some minor map edits to make the surf area accessible. See the demo below.

Cherrygrove Nugget


Rangi42:
Nice work! Red++ 4 already did this (https://github.com/Rangi42/RedPlusPlus/tree/master/notes/map%20stitches), but I never got around to porting it to PC.

PR #676, opened 2022-01-29
Add the hidden items from HGSS on Route 32

rik-smeets:
See `missing-hgss-items.txt` for the changes. I also removed the section on Violet City since that Pokéball was already a hidden item there.


Rangi42:
Thanks for this! Note that I'm not committed to using all of these HGSS items, and some could be adapted to different locations. For instance, PC doesn't have Rock Climb, but it does have new areas like Route 35 Coast, so an item like "Rare Candy	Near the north-east corner of the Lighthouse (requires Rock Climb)" could be hidden there instead.

PR #677, opened 2022-01-29
Add hidden nugget from HGSS on route 35

rik-smeets:


Rangi42:
Thanks again! I edited this to append new event constants to the very end of the event flags. It's less well-organized, but avoids breaking saves.

rik-smeets:
@Rangi42 I wasn't aware of that, thanks! I'll keep that in mind in upcoming PR's.

PR #680, opened 2022-02-18
Correct the Butterfree dex entry

fosskers:
Bees make honey from flowers, but the flowers themselves only emit pollen which many bugs feast on.


itsdarsh:
This isn't entirely true, as plants also produce nectar to attract bugs, and nectar is the sugar source of honey. I'd be willing to entertain changing references of honey to nectar, but this is something from vanilla and honey is more accurate than pollen in this case (even if both are wrong).

PR #683, opened 2022-03-20
Fix Sylveon cry in intro

JanJastrow:
Hi, while testing 9bit I noticed that the text at this position auto-scrolled by and no Pokémon cry was audible.
Judging by pokecrystal, this should be a ```text_end``` ?
https://github.com/pret/pokecrystal/blob/master/data/text/common_2.asm#L1544


Rangi42:
Thanks for catching this bug!

PR #687, opened 2022-04-19
Use menu_coords macro for menu headers

nickjwilde:
Looking to contribute where I can.


Rangi42:
Thanks! This is definitely an improvement.

Issue #681, opened 2022-02-21
Miscellaneous 9bit issues


itsdarsh:
- [x] Alolan Raichu's front sprite animation is bugged. (https://discord.com/channels/332698009060114434/933019563535253534/934551623437525012)
- [x] Celadon Eusine and Bill take a very long time to check if the player has the legendary beasts, delaying opening their textboxes significantly.
- [x] Certain babies cannot be hatched. Mantyke, Tyrogue, and possibly others. You get the evolved form instead.
- [x] Gender display is messed up when using items like Rare Candy. (https://discord.com/channels/332698009060114434/933019563535253534/933380995120525322)
- [x] Occasionally, mons will come out of eggs with the shiny palette and then not be shiny when you check them. (https://discord.com/channels/332698009060114434/933019563535253534/933356508413054986)
- [x] The Hall of Fame sequence shows incorrect species names. (https://discord.com/channels/332698009060114434/933019563535253534/933347429535870986)
- [x] Unown counter isn't working
- [x] Cancelling an evolution crashes the game.
- [x] Odd Souvenir says not able for (Surfing only?) Pikachu but can be used on it regardless.
- [x] Refreshing the battle screen by going into a menu and back shows normal Pikachu even if the foe is Surfing/Flying.
- [x] Leveling up and triggering Pickup at the end of battle messes up the palettes. (https://discord.com/channels/332698009060114434/933019563535253534/933028014848106616)
- [x] Pokémon sent out by enemy trainers all have the same cry. **(Cannot reproduce)**
- [ ] Moves that trigger additional messages will basically skip the "X used Y!" text on Instant Text. **(Not specific to 9bit)**
- [x] Abilities like Static that modify encounter chances don't seem to be working correctly.
- [x] Battle Factory rentals are broken
- [x] Clock is reset to midnight when starting a new game

Issue #567, opened 2020-08-24
engine/pokemon/search.asm needs refactoring to handle flags properly


earthoul:
> `#questions`
>
> 2020-08-23 20:45 UTC
> Rangi: but I may just replace the whole logic with flags for when you catch \[legendaries\]
>
> 2020-08-23 20:53 UTC
> Rangi: \[…\] There are too many possible combinations of "triggered the event, seen the mon, caught the mon, still own the mon".  I'll bet this is responsible for \[#208\].

[search.asm](https://github.com/Rangi42/polishedcrystal/blob/feb69f82570f9b010ce77f931c63a8941035f3f4/engine/pokemon/search.asm) contains `SpecialBeastsCheck`, `SpecialBirdsCheck`, `SpecialDuoCheck` and more, which are also related to some flags such as Eusine and Lawrence. Some cases (not fully descriptive yet, though), including but not limited to #208, don't set those flags properly.

Found by: `@Salamandar`, `@singularity`, `@Rangi` & Co. on `#questions`

Rangi42:
Related pokecrystal bug: https://github.com/pret/pokecrystal/blob/master/docs/bugs_and_glitches.md#checkownmonanywhere-does-not-check-the-day-care

PR #690, opened 2022-06-02
Search ASM refactor

vulcandth:
Okay; so the following was an idea I had in refactoring search.asm #567 . Feel free to use some, all, or none of this. One of the things I noticed as being a problem is the reliance on Pokedex caught flags to determine if a legendary pokemon has been caught. The problem with using these is the fact it doesn't verify that the legendary was caught by the Player. From what I understand (unless polishedcrystal changes this somewhere), traded pokemon, for example, would set the pokedex caught flag. Also, reliance on `CheckOwnMonAnywhere` has a problem where the user could trade away their Legendary, hide it somewhere (lookin at you daycare), or something else.

Changes:

- Added `ENGINE_PLAYER_CAUGHT_*` engine flags to record the _**Player**_ has caught the mon. These are currently only used for the one-offs.
- `wBattleResult` now always sets bit 6 if pokemon is caught.
- Repurposed `CheckCaughtCelebi` to `CheckBattleCaughtResult`. It now checks `wBattleResult` if pokemon was caught
- Added `BATTLERESULT_*` constants and applied throughout code.
- Simplified `SpecialBeastsCheck` `SpecialBirdsCheck`, and `SpecialDuoCheck`
- Removed `SpecialMonCheck` and `CheckOwnMonAnywhere`

Todo:

- [x] - Record Sudowoodo caught result `wBattleResult` (Currently only `BATTLETYPE_LEGENDARY` and `BATTLETYPE_ROAMING`)
- [x] - Test capturing SUDOWOODO sets the correct `ENGINE_PLAYER_CAUGHT_SUDOWOODO` bit in `wPlayerCaught2`
- [x] - Test capturing SUICUNE sets the correct `ENGINE_PLAYER_CAUGHT_SUICUNE` bit in `wPlayerCaught` (also tests `CheckBattleCaughtResult`)
- [x] - Test **failing** to capture SUICUNE correctly does not set the `ENGINE_PLAYER_CAUGHT_SUICUNE` flag.
- [x] - Test capturing roaming RAIKOU and ENTEI sets their respective `ENGINE_PLAYER_CAUGHT_*` flags.
- [x] - Test RAIKOU and ENTEI appear in Tin Tower if they haven't been caught.
- [x] - Test RAIKOU and ENTEI do not appear in Tin Tower if they were successfully caught prior.
- [x] - Test EUSINE in `EUSINES_HOUSE` Skips beasts dialog and showing the player beats if all beasts have been caught. (This tests `SpecialBeastsCheck`)
- [x] - Test EUSINE in `EUSINES_HOUSE` Skips showing individual beasts if they have already been caught by the player.
- [x] - Test `RespawnOneOffs` Specially Suicune Roaming if wasn't caught.
- [x] - Test `RespawnOneOffs` if Entei and Raikou were caught. Ensured that they were not respawned.
- [ ] - Other tests

Possible further improvements:

- [ ] - Rework more of the event scripts to utilized these flags instead of others.

Possible alternatives to my implementation:

- Remove `CheckOwnMonAnywhere` and `SpecialMonCheck` only and utilize standard Event flags to keep track of things.

Final Note:

Yes, I put some time in this.. Mainly this has been something to help myself learn. I will have no hard feelings if you don't think this is the correct approach and close this PR. I will open as a draft for now, Feel free to look it over and let me know if I need to not waste any more time on this.

References:
#681


engine/events/specials.asm
```diff
@@ -354,54 +354,78 @@ RespawnOneOffs:
 	eventflagreset EVENT_BEAT_CYNTHIA
 .SkipCynthia
 
-	ld a, SUDOWOODO - 1
-	call CheckCaughtMon
+	ld de, ENGINE_PLAYER_CAUGHT_SUDOWOODO
+	ld b, 2 ; check
```
Rangi42, line 358: There's a `CHECK_FLAG` constant for this.

engine/pokemon/search.asm
```diff
@@ -1,81 +1,35 @@
...
-
-	ld a, SUICUNE
-	ldh [hScriptVar], a
-	call CheckOwnMonAnywhere
-	jr nc, SpecialMonCheckFailed
-
-	; they exist
+	ld a, [wPlayerCaught]
+	and CAUGHT_BEASTS_MASK
+	cp CAUGHT_BEASTS_MASK
```
Rangi42, line 7: https://github.com/pret/pokecrystal/wiki/Optimizing-assembly-code#compare-a-to-a-mask-after-masking-it

engine/pokemon/search.asm
```diff
@@ -1,81 +1,35 @@
...
-
-	ld a, MOLTRES
-	ldh [hScriptVar], a
-	call CheckOwnMonAnywhere
-	jr nc, SpecialMonCheckFailed
-
-	; they exist
+	ld a, [wPlayerCaught]
+	and CAUGHT_BIRDS_MASK
+	cp CAUGHT_BIRDS_MASK
```
Rangi42, line 19: https://github.com/pret/pokecrystal/wiki/Optimizing-assembly-code#compare-a-to-a-mask-after-masking-it

engine/pokemon/search.asm
```diff
@@ -1,81 +1,35 @@
...
+; Check if the player has caught Lugia and Ho-Oh.
 ; Return the result in hScriptVar.
 
-	call CheckOwnMonAnywhere
-	jr nc, SpecialMonCheckFailed
-
-	; they exist
+	ld a, [wPlayerCaught]
+	and CAUGHT_DUO_MASK
+	cp CAUGHT_DUO_MASK
```
Rangi42, line 31: https://github.com/pret/pokecrystal/wiki/Optimizing-assembly-code#compare-a-to-a-mask-after-masking-it

engine/pokemon/search.asm
```diff
@@ -85,68 +39,15 @@ SpecialMonCheckFailed:
...
-	scf
-	ret z
-	jr .cmp_ot
-.next
-	dec c
-	jr nz, .loop
-	dec b
-	bit 7, b ; check for reaching -1
-	jr z, .outer_loop
+CheckBattleCaughtResult:
```
Rangi42, line 42:
```
CheckBattleCaughtResult:
	ld a, [wBattleResult]
	and 1 << 6
	rlca
	rlca
	ldh [hScriptVar], a
	ret
```

engine/events/specials.asm
```diff
@@ -354,54 +354,78 @@ RespawnOneOffs:
 	eventflagreset EVENT_BEAT_CYNTHIA
 .SkipCynthia
 
-	ld a, SUDOWOODO - 1
-	call CheckCaughtMon
+	ld de, ENGINE_PLAYER_CAUGHT_SUDOWOODO
+	ld b, CHECK_FLAG
+	call _EngineFlagAction
+	ld a, c
+	and a
```
vulcandth, line 361: A lot of redundant code here... Not sure if I should optimize for cycles or bytes here.

Optimizing for bytes would be something like:
```asm
	ld de, ENGINE_PLAYERCAUGHT_SUDOWOODO
	call CheckCaughtFlag
	jr nz, .CaughtSudowoodo
	eventflagreset EVENT_ROUTE_36_SUDOWOODO
```

```asm
CheckCaughtFlag:
	ld b, CHECK_FLAG
	call _EngineFlagAction
	ld a, c
	and a
	ret
```
I'm estimating this would save roughly 33 bytes... but add an extra 110 cycles.

Rangi42: The whole routine is doing a lot of slow flag work anyway, so may as well optimize for size.

vulcandth:
Please note that merging this will cause small problem with old saves. If the player has caught any of the one-offs already, the respawnoneoff code will respawn all of them again. As old saves will not have any of the caught flags set.

vulcandth:
Alright I think i'm ready for official review. Currently this is targeting Master; let me know if you want it rebased for 9-bit instead. Also please let me know if you want other tests performed. I'll do any you feel is necessary, see the first post on the list of tests i've already performed.

engine/events/specials.asm
```diff
@@ -435,11 +435,10 @@ RespawnOneOffs:
 	ret
 
 CheckCaughtFlag:
-; Sets the z flag if the player has not caught the mon
+; EngineFlagAction sets the z flag if the player has not caught the mon
 ; input: de = ENGINE_PLAYER_CAUGHT_*
 	ld b, CHECK_FLAG
-	call _EngineFlagAction
-	dec c
+	farcall EngineFlagAction
```
Rangi42, line 441: You should always turn `call Foo / ret` into `jmp Foo`, and `farcall Foo / ret` into `farjp Foo`. (PC uses `jmp` as a macro to warn when a `jp` can be optimized to `jr`.) Run `utils/optimize.py` to double-check such micro-issues.

constants/battle_constants.asm
```diff
@@ -311,6 +311,10 @@ QUICK_PACK   EQU 1 << QUICK_PACK_F
 	const LOSE
 	const DRAW
 
+DEF BATTLERESULT_CAUGHT_POKEMON EQU 6
```
Rangi42, line 314: Generally bitflag constants end in `_F`.

Rangi42: Also `const_def` would be used even for such a short sequence:
```
const_def 6
const BATTLERESULT_CAUGHT_POKEMON  ; 6
const BATTLERESULT_BOX_FULL        ; 7
```

engine/items/item_effects.asm
```diff
@@ -499,11 +499,11 @@ PokeBallEffect:
 	ld a, [wBattleType]
 	cp BATTLETYPE_CONTEST
 	jmp z, .catch_bug_contest_mon
-	cp BATTLETYPE_LEGENDARY
-	jr nz, .not_celebi ; false positive for other legendaries, but that's okay
+
+	; Set wBattleResult bit 6 to record
+	; that the pokémon was caught.
```
Rangi42, line 504: `set BATTLERESULT_CAUGHT_POKEMON, [hl]` is obvious enough that this comment can be removed. Especially since, if the value of `BATTLERESULT_CAUGHT_POKEMON` were ever changed from 6, the comment would become misleading.

constants/battle_constants.asm
vulcandth: This section was actually just lifted from pokecrystal.. but easy enough.

engine/events/specials.asm
vulcandth: The more I read this; the more confused I get. I understand that `jmp` should be used instead of `jp` because it can provide an assert warning if it can be optimized to `jr`, but how can you replace a `call Foo / ret`? Aren't `jp`/`jr` used if you don't need to `ret`?

And looking at the `farjp` macro; i'm not sure what the significance of `BANK(\1) | $80` is doing compared to a normal `farcall`. 

Oh, and I ran utils/optimize.py without changing anything.. and it doesn't appear to have found anything wrong with what i'm doing.

I'm not being argumentative here... I love these learning opportunities for myself!

ctuncan: If the only thing you do after a call is a ret, then it can be replaced by a jmp. This is known as a "tail call" if you wanna look up more details on it.

But here's my attempted explanation:
Call sets up a "stack frame" (in general computer science. Here the only thing in the frame is the return address)
Ret then takes the latest stack frame and returns to that location.

So for example:

```
start:
  call A // 1. Set up stack frame #1 and jump to A
  nop // Stack Frame #1 points here. 5. We get here!
  // END
  
A:
  call B // 2. Set up stack frame #2 and jump to B
  ret // Stack frame #2 points here. 4. pop stack frame #1 and jump to start's nop
  
B: 
  ret // 3. Pop stack frame #2 and jump to A's ret
```

I tried to label what happens at each instruction here, in numbered order, hope you can follow!
The tail call optimisation says that stack frame #2 is useless! Why not just jump straight from B back to start, as we don't need to do anything more at A.

Since all call does is setup the stack and jump, we can just jump to skip setting up the stack.
And because we jump out of A, we don't need to ret any more either as we would never reach that instruction.

Sorry if I'm stepping on your toes in your own repo Rangi!

ctuncan: Essentially consider `call` == `push pc; jp`, and `farcall `== `push pc, farjp`, while `ret` == `pop pc`.

This means that an unoptimised tailcall is equivalent to:
```
push pc
jp
push pc
jp
//do work
pop pc 
pop pc
```

The inner push/pop isn't needed as you never use the value you're saving.
The same applies if you push/poping registers that don't actually need to be saved.

Rangi42 (approved):

PR #703, opened 2022-07-09
Swap ABILITY_1 and ABILITY_2 in base data

etdv-thevoid:
These changes address the following comment noted in `data/trainers/parties.asm`:

> TODO: swap ABILITY_1 and ABILITY_2 in base data as need so that ABILITY_1 is the better one for default NPCs (e.g. NPCs should have Sturdy when possible)


Rangi42:
Thanks, but that's outdated. The recent plan is to set ability 1 or 2 for plain NPC trainer mons, in such a way that they're roughly evenly distributed, in a deterministic way (no Random calls). Example: hash the NPC's name and the mon's species together.

PR #709, opened 2022-07-20
Fix transparency on Pokeballs

Dev727:
As to not mess up the animation constants, I added the new one and everything associated with it to the end of all the lists.
This also utilizes the GREEN palette and uses the Pokeball RGB values to keep a pure white image.

One thing to note is that the Master Ball animation currently is reduced to 4 sparkles as 8 will be 2 too many.

I will eventually look into adjusting the animation itself and get a 6 sparkle variant working.


Rangi42 (approved):
This looks good, thank you!

Rangi42:
It doesn't seem to actually do anything: Poké Balls still appear transparent, there are no new bg sprites in the debugger.

Be sure to test with a two-color ball like Great/Ultra/Master, since Poké Balls already have opaque white as one of their two colors.

image

image

PR #715, opened 2022-07-21
Fix GFX Value

Dev727:
I can't count, and I forgot to pull updates earlier...

This should fix it though.


Rangi42:
Thanks again! PC should switch to using GFX constants for those.

PR #716, opened 2022-07-21
Swap to GFX constants

Dev727:
Swapped objects.asm from values to constants.

Also, changed the copy palette in helpers.asm to directly load the white value from CaughtBallPals.

The copy method was just duping the ball color rather than using white.


Rangi42:
Thank you, but I did this in the `9bit` branch: https://github.com/Rangi42/polishedcrystal/commit/97bb3f46719c9d9776869be4581eecf5ce4a0cf5 (it's where most of the work is going on, and will be merged into `master` Soon)

PR #719, opened 2022-07-26
Workflow overhaul

etdv-thevoid:
Seperates main.yml into build.yml and nightly.yml

This change allows build.yml to be called from Pull Requests, adding an additional check towards "completed" status so that an automated process can ensure that PRs will build after merge.


etdv-thevoid:
This now addresses Issue #726 by including a `nortc` build in the process as well.

PR #735, opened 2022-08-21
Update Chuck's teams to use Galarian Farfetch'd & Sirfetch'd

etdv-thevoid:
The 9bit update includes the addition of Galarian Farfetch'd & Sirfetch'd, with plain form Farfetch'd changed back to its vanilla typing. Yet Chuck's team is not changed to reflect this.

This corrects that by giving Chuck Galarian Farfetch'd in his initial team & Sirfetch'd in the rematch. Replacing Machoke / Machamp, this also further differentiates Chuck's team from Bruno's.

---

Also included in this PR is the swapping of plain form Farfetch'd for Galarian Farfetch'd on Routes 47 & 48 (as per @Rangi42's suggestion).


FredrIQ:
For trainer parties when setting abilities, please use ABILITY_(species)_(ability). For example ABILITY_FERALIGATR_TORRENT.

I know the parties file currently doesn't follow this structure, but that's because nobody bothered addressing it. I think @itsdarsh intended to?

PR #739, opened 2022-08-25
Disable changing the ball of a traded pokémon

wolfboyft:
...Unless the traded-pokémon-treat-you-as-OT option is set.


wolfboyft:
Yay! Thanks for the merge :-)

PR #741, opened 2022-08-25
Update three water-type move animations

etdv-thevoid:
Updates the following three water-type move animations that each have rather lackluster vanilla animations in Gen II:

- Crabhammer
- Waterfall
- Octazooka

Each one now has the simple whirlpool bgeffect applied to simulate the water effects used with these moves in later generations.

Builds upon the work of FIQ and SoupPotato in their series of animation updates. See PR #737

### DIFF
--- a/data/moves/animations.asm
+++ b/data/moves/animations.asm
@@ -2675,6 +2675,7 @@ BattleAnim_Swift:
 
 BattleAnim_Crabhammer:
 	anim_1gfx ANIM_GFX_HIT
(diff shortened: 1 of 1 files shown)


Rangi42:
Thanks, but I don't think a full-screen bg effect is appropriate for these moves. For example, Gen 4 Crabhammer shows show water drops along with the impact hit, and Octazooka shows smoke. Waterfall has a full-screen water bg that could be devamped as the Whirlpool effect, but just droplets along with the impact could also work.

PR #761, opened 2022-12-10
Phone/Oakratings refactor

vulcandth:
Refactors the Oak ratings and Phone data to be more in line with Pokecrystal, as well as provides some relabeling. Verified that these changes do not change the sha1 of the rom.


engine/phone/scripts/jack_gossip.asm
```diff
@@ -0,0 +1,68 @@
...
+
+.Gust:
+	farwritetext JackGustTriviaText
+	promptbutton
+	sjump PhoneScript_HangupText_Male
+
+.Twister:
+	farwritetext JackTwisterTriviaText
+	promptbutton
+	sjump PhoneScript_HangupText_Male
```
vulcandth, line 43: These labels should be relabeled since Twister is not a move in 9-bit. Need recommendations on what you would like. You may also want to revisit the actual text as well.

PR #763, opened 2022-12-17
Sync `OBJECT_` constant names with pret

vulcandth:
Sync's the `OBJECT_` constant names with pret. I used REGEX to do this, so if you prefer your current constant names, feel free to close this. I was trying to read through some polished code, and the OBJECT names conflicting with pret's was throwing me off.. I'm sure it would for others if you plan for others to use polished as a base. 

I verified the sha1 of the rom is not changed with this PR.


Rangi42:
This is a complex subsystem, so thank you, it's helpful to agree with pret there!

PR #764, opened 2022-12-17
Misc pret commits

vulcandth:
Couple more misc pret commits I had time to pull over and removal of some whitespace.

does not change sha1.


engine/battle/effect_commands.asm
```diff
@@ -4483,7 +4483,7 @@ BattleCommand_sleep:
 
 	; Check if we were cured
 	ld a, BATTLE_VARS_STATUS_OPP
-	cp 1 << SLP
+	cp 1 << SLP_MASK
```
Rangi42, line 4486: @FredrIQ This is a suspicious check; isn't it the same as `1 << TOX`?

PR #769, opened 2022-12-22
Full Party Naming Screen Issues

vulcandth:
I've only managed to partially fix this so far...

The issue mainly was:

If a party is "full", `LoadNamingScreenMonMini:` would call `GetPartyParamLocationAndValue:` and use `wPartyMon1Form` instead of `wTempMonForm`. Thus, if an Egg Or a pokemon with a Form was in the first slot, and the player catches a pokemon.. The naming screen will either show an Egg icon/text or a glitched icon.

This is only a partial fix so far... Although the correct icon/text now loads.. the palette is still wrong. I haven't been able to figure that out yet...

### DIFF
--- a/engine/gfx/cgb_layouts.asm
+++ b/engine/gfx/cgb_layouts.asm
@@ -595,6 +595,12 @@ _CGB_NamingScreen:
 	rst AddNTimes
 	ld d, h
 	ld e, l
+	ld a, [wMonType]
+	cp TEMPMON
+	jr nz, .party_mon
+	call LoadTempMonPalette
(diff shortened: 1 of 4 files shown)


vulcandth:
Fixed the palettes as well. Tested and verified it is working. Also got it working when using the `givepoke` mapscript command.

vulcandth:
This should resolve the following reported issues:

https://discord.com/channels/332698009060114434/933019563535253534/1056869128079687771
https://discord.com/channels/332698009060114434/933019563535253534/1054574652908908645

PR #772, opened 2023-01-02
9bit

Lethaldragon1988:
Please consider giving Furret a buff, Furret is such an iconic Pokemon. Consider slightly increasing attack and buffing defences. Perhaps this will require an increase to evolution level. I love your hack more than any other Pokemon hack :)


Rangi42:
Looks like something went wrong creating this PR; it's trying to merge PC's own 9bit branch into master, there's no commit with your Furrett changes. That said, I think its +28-buffed 443 BST is fine for a first-route Pokémon. Enjoy whatever .asm changes you want in your own copy though!

PR #781, opened 2023-01-14
Update ThunderWaveDescription

rjd1922:
Change Thunder Wave description from "A move that may cause paralysis." to "A move that causes paralysis." since it always paralyzes.

### DIFF
--- a/data/moves/descriptions.asm
+++ b/data/moves/descriptions.asm
@@ -478,6 +478,9 @@ WillOWispDescription:
 
 StunSporeDescription:
 ThunderWaveDescription:
+	db   "A move that"
+	next "causes paralysis.@"
+
 GlareDescription:
 	db   "A move that may"
 	next "paralyze the foe.@"


data/moves/descriptions.asm
```diff
@@ -478,6 +478,9 @@ WillOWispDescription:
 
 StunSporeDescription:
 ThunderWaveDescription:
+	db   "A move that"
+	next "causes paralysis.@"
+
 GlareDescription:
```
Rangi42, line 484: Shouldn't Glare also be updated?

rjd1922: Sorry it took me forever to see this. Yes, Glare should probably have the same description as Thunder Wave saying it always paralyzes (it it hits; Glare only has 75% accuracy.

rjd1922: Actually, based on how Hypnosis's description says it "MAY cause the foe to fall asleep," while Spore says it "PUTS the foe to sleep," I assume moves with 100% accuracy are supposed to say they "may" cause a status effect.

rjd1922: I made a new pull request that includes more changes.

itsdarsh:
Closing this just because it's redundant with a larger pr.

PR #791, opened 2023-02-02
Allow Heal Bell to Trigger Throat Spray Buff

f-raZ0R:
Heal Bell, as a sound based move in the mainline games, should be able to trigger throat spray's +1 SpA buff when used by a pokemon holding the item. As heal bell was not currently in the db for sound moves, it currently does not do so. This PR aims to fix that.

Possible Issues:
-This might interact weirdly if a pokemon with soundproof uses heal bell? The vanilla games already have a somewhat weird soundproof+heal bell interaction anyways

### DIFF
--- a/data/moves/sound_moves.asm
+++ b/data/moves/sound_moves.asm
@@ -8,4 +8,5 @@ SoundMoves::
 	db SCREECH
 	db SING
 	db SUPERSONIC
+	db HEAL_BELL
 	db -1


FredrIQ:
Does it work? I know that non-attacking sound moves is missing part of the sound move logic (it's a todo item on my end).

f-raZ0R:
Unfortunately it does not currently function, likely due to the aforementioned missing logic you refer to.

PR #803, opened 2023-03-02
Some optimization

ariscop:
Stuff i've been working on, mostly against base crystal, but would like to see used somewhere; Much of my changes don't apply to polished but these few do

And a random fix to WaitPressAorB, so it doesn't busy the cpu when stat changes on level up are shown

UpdateBGMapBuffer now depends on  wBGMapBuffer and wBGMapPalBuffer not crossing a 256 byte boundary, they don't but not sure how i should assert for that

Hopefully Decompress is readable enough. Subroutines are in an... eclectic order, mostly from rearranging whenever they were out of jr range, and the code quite dense, some of which is unavoidable

### DIFF
--- a/home/copy.asm
+++ b/home/copy.asm
@@ -12,32 +12,71 @@ FarCopyBytes::
 _CopyBytes::
 ; copy bc bytes from hl to de
 	inc b  ; we bail the moment b hits 0, so include the last run
-	inc c  ; same thing; include last byte
-	jr .HandleLoop
-.CopyByte:
+
+	srl c
+	jr nc, :+
 	ld a, [hli]
 	ld [de], a
 	inc de
-.HandleLoop:
+
+:	srl c
+	jr nc, :+
+	ld a, [hli]
+	ld [de], a
+	inc de
+	ld a, [hli]
+	ld [de], a
+	inc de
+
+:	jr z, :++
+:	ld a, [hli]
+	ld [de], a
+	inc de
+	ld a, [hli]
+	ld [de], a
+	inc de
+	ld a, [hli]
+	ld [de], a
+	inc de
+	ld a, [hli]
+	ld [de], a
+	inc de
+
 	dec c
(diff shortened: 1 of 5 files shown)


FredrIQ:
I actually did consider unrolling CopyBytes/ByteFill, but to every 16th instead of every 4th since I figured it'd simplify the counting (considering the swap instruction). But even the smaller unroll should be a rather significant boost.

Anyway, the way you changed WaitPressAorB introduces a small issue: a very poorly timed vblank can cause it to drop a button input.
Consider this:

	call CheckIfAOrBPressed
	jr nz, .done
	; vblank where a pending A press happened
	call DelayFrame ; this will forget about the A press
	jr .loop

I have considered adding a MaybeDelayFrame conditional to resolve this sort of problem: basically set a flag before calling CheckIfAOrBPressed, "pending delay" or similar and MaybeDelayFrame would only DelayFrame if the flag remains set (and if LY isn't close enough to vblank to avoid the race condition, otherwise busyloop until vblank). This is a negligible issue in this particular situation with CheckAOrBPressed, but there are other situatios where I *really* don't want a badly timed vblank to throw things off, cause double delays or whatnot. But maybe this is overkill?

Either way, thanks for the PR! I see no reason not to merge most, if not all, of it, just wondering your thoughts on the above.

ariscop:
I don't think vblank can break it, hJoypadDown is updated by vblank, GetJoypad reads that once and compares against hJoyDown, which is updated by GetJoypad itself, making the interaction atomic

Also there are both hJoy* and hJoypad* variables which is news to me despite having looked at this code before

For CopyBytes/ByteFill 4 iterations is mostly about size, since it grows exponentially; Also tried 8 but it didnt make the continue button noticeably faster
Wrote a macro for the unrolling if you want to try other counts https://gist.github.com/ariscop/f067431e133c34a6308e10a90379c7ed

PR #806, opened 2023-03-05
Some Dynamic Pal Improvements

vulcandth:
* No longer run Dynamic Pal in `LoadMapPals`
* Don't update `OBJECT_PALETTE` during the first dyn pal pass.
* Better Menu exit hook (Doesn't corrupt BG)
* Ensure the correct Wram bank is loaded.
* Preserve more registers, just in case.
* Re-fix Battle Transitions
* Re-fix sprites appearing before warp is complete
* Fix sprite problems when activating surf (Menu problem).
* Verified pals updating correctly when crossing map connections with different map conditions (Example: Overcast)

### DIFF
--- a/engine/gfx/dynamic_pals.asm
+++ b/engine/gfx/dynamic_pals.asm
@@ -42,7 +42,7 @@ CheckForUsedObjPals::
 
 	ld hl, wPalFlags
 	bit DISABLE_DYN_PAL_F, [hl]
-	jmp nz, .done
+	jr nz, .done
 
 	; reset all wUsedObjectPals bits
 	xor a
@@ -66,7 +66,7 @@ CheckForUsedObjPals::
 
 .done
 	pop af
-	ldh a, [rSVBK]
+	ldh [rSVBK], a
 	jmp PopAFBCDEHL
 
 ScanObjectStructPals:
--- a/engine/gfx/dynamic_pals.asm
+++ b/engine/gfx/dynamic_pals.asm
@@ -1,20 +1,32 @@
 ClearSavedObjPals::
+	ldh a, [rSVBK]
+	push af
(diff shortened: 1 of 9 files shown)


engine/gfx/dynamic_pals.asm
```diff
@@ -1,20 +1,32 @@
...
 
 DisableDynPalUpdates::
+	push hl
 	ld hl, wPalFlags
 	set DISABLE_DYN_PAL_F, [hl]
+	pop hl
 	ret
 
 EnableDynPalUpdates::
+	push hl
```
Rangi42, line 26: `CheckForUsedObjPals` doesn't preserve `hl`, so either this is unnecessary or `CheckForUsedObjPals` should do it too.

home/map.asm
```diff
@@ -1640,6 +1640,8 @@ ReturnToMapWithSpeechTextbox::
 	ld [wSpriteUpdatesEnabled], a
 	call ClearBGPalettes
 	call ClearSprites
+	farcall ClearSavedObjPals
```
Rangi42, line 1643: After FIQ and ariscop were discussing `farcall`s, I wonder if this adds any perceptible delay to the overworld textboxes.

engine/gfx/dynamic_pals.asm
Rangi42: `CheckForUsedObjPals` already does. <_<

home/map.asm
ariscop: Perhaps the discussion in discord has overstated the problem; they're on the order of 10x the overhead of a plain call/ret, which isn't that bad in practice

ReloadTilesetAndPalettes already does a lot of farcalls, a few more is unlikely to be noticeable

PR #807, opened 2023-03-05
Big farcall refactor

ariscop:
Stub farcalls in home, a artefact from earlier in the codes history, obscured the extent of farcall usage and allowed useless farcalls to escape detection

GetTrueUserAbility and GetOpponentAbilityAfterMoldBreaker were the bigest offenders, with almost half the call sites being from their own bank, but silently farcalling due to stubs in home

Outside of those there's no obvious issues but this has allowed the removal of some farcalls through changes to bank allocation

Some changes to the script engine were also made in hopes of allowing more overworld related code to live in the same bank, as well as relocating 3 bytes to hram

### DIFF
--- a/engine/battle/abilities.asm
+++ b/engine/battle/abilities.asm
@@ -15,7 +15,7 @@ RunEnemyStatusHealAbilities:
 RunStatusHealAbilities:
 	ld hl, StatusHealAbilities
 UserAbilityJumptable:
-	call GetTrueUserAbility
+	farcall GetTrueUserAbility
 AbilityJumptable:
 	; If we at some point make the AI learn abilities, keep this.
 	; For now it just jumps to the general jumptable function
@@ -253,7 +253,7 @@ IntimidateAbility:
 	ld a, [wFailedMessage]
 	and a
 	jr nz, .continue
-	call GetTrueUserAbility
+	farcall GetTrueUserAbility
 	cp RATTLED
 	ld b, SPEED
 	call z, StatUpAbility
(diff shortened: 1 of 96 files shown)


Rangi42:
So first of all, thanks for doing this! Long-term reducing farcalls is a good goal.

However, is this PR mainly because the bank 0 stubs "obscured the extent of farcall usage and allowed useless farcalls to escape detection"? If so, I'd prefer not to merge it. The stubs exist to save ROM space, which I expect why this PR had to introduce more sections.

How about changing their labels? That would also let you diff the ROM before and after, and it should be identical.

```diff
-LoadFrame::
 -	farjp _LoadFrame
+FarStub_LoadFrame::
+	farjp LoadFrame
```

Rangi42:
> Some changes to the script engine were also made in hopes of allowing more overworld related code to live in the same bank, as well as relocating 3 bytes to hram

If these are saving space and/or cycles on their own, they can be their own PR.

ariscop:
> However, is this PR mainly because the bank 0 stubs "obscured the extent of farcall usage and allowed useless farcalls to escape detection"? If so, I'd prefer not to merge it. The stubs exist to save ROM space, which I expect is why this PR had to introduce more sections.

By obscure i mean the macros and farcheck.py can't see through the stub, and it's difficult to know which files share a section
In terms of space it's barely larger, 4 bytes for a farcall/farjp vs 3 for an absolute call/jp, additional sections was about bringing related code together rather than space limits


> If these are saving space and/or cycles on their own, they can be their own PR.

Will do

ariscop:
Going to do these changes in smaller chunks, plus this branch is kind of stale by now

PR #809, opened 2023-03-05
Remove ROM0 wrappers for farcalled math and trig functions

ariscop:
Moves math and sine functions to their own sections, placed in the same bank as some of their callers. doing so trades about 160 romx bytes 30 or so rom0 bytes

Very small optimization of moving RstFarJump to high home, inside jr range


home/farcall.asm
```diff
@@ -17,52 +17,6 @@ MACRO update_pushed_af_flags
 	pop af
 ENDM
 
-FarCall_de::
```
Rangi42, line 20: Why rearrange this? Did it enable a particular `jp`→`jr` somewhere?

layout.link
```diff
@@ -82,6 +82,7 @@ ROMX $0E
 
 ROMX $0F
 	"Battle Core"
+	"Math"
```
Rangi42, line 85: Do these sections need to be in these particular banks?

Rangi42:
I'm just not sure what this ultimately gains. It's a net loss of ROM space, although that doesn't really matter; 160 ROMX bytes are affordable, but we weren't needing 30 more ROM0 bytes either. This makes each Multiply/Divide/Sine/Cosine call site a few cycles faster and one byte larger. Is the eventual plan to group the math code with other code so that many of these can become regular calls?

ariscop:
> Is the eventual plan to group the math code with other code so that many of these can become regular calls?

That's why its placed with battle core, being one of the largest users of multiply/divide. similarly sine was placed in bank 23 next to the sprite animation code; Every remaining call represents a removed farcall

It's not needed now but freeing ROM0 bytes creates opportunity. sine could be moved in and would reduce overall rom usage for example, though i would imagine there's higher value changes

Rangi42:
Fair enough. It would be podsible later on to add back ROM0 wrappers for frequent farcalls if it's ever desirable, like how Ithe most-called functions were converted to rsts.

PR #816, opened 2023-03-09
Save registers on stack instead of hram during farcalls

ariscop:
Making them re-entrant which allows safe use in interrupts, removing a potential source of hard to diagnose bugs; Probably still shouldn't be used in interrupts but it won't be a source of bugs now

Larger due to different stack layouts making it not possible to reuse code, though it is reused for returns where possible


ariscop:
Not entirely happy with the stack call functions, they could be simplified dramatically if the caller pushed something to the stack first

Not important what they push, its just the cheapest way to bump the stack pointer

Rangi42:
So, this increases ROM usage; how does it affect farcall cycles?

I don't see any value in making farcall safe(r) in interrupts, since that can easily just never be done.

ariscop:
Didn't check performance after confirming that farcall itself was faster, and the rest either should be faster or comparable

Though interrupts shouldn't use farcall, they can use more effecient methods, it could happen unintentionally and would cause difficult to reproduce bugs. Removing the use of global state eliminates that entirely

Rangi42:
> farcall itself was faster, and the rest either should be faster or comparable

alright

Rangi42:
@ariscop  I merged this before seeing that #817 said it "needs more work", so I force-pushed to undo the merge.

PR #818, opened 2023-03-10
DynPal Fix Flash in Darkness and `FadeInPalettes`

vulcandth:
Fixes objects going white when using FLASH in darkness. This also fixes `FadeInPalettes` when used in various map scripts.


vulcandth:
Retracting this for a moment... Realized I needed to check something... sorry.

Rangi42:
So what's the guideline for when to use a `special EnableDynPalUpdatesNoApply` when scripting?

PR #819, opened 2023-03-10
Save registers on stack instead of hram during farcalls

ariscop:
Farcalls are slightly slower (146 vs 140 cycles) but re-entrant, thus safe to use in interrupts,  and 4 hram bytes have been freed

Going to let this sit as a draft, make sure i haven't missed something like the last pr


home/header.asm
```diff
@@ -9,17 +9,10 @@ EntryPoint::
...
-PopBCDEHL::
-	pop bc
-	pop de
-	pop hl
-	ret
-
+FarCall::
+	dec sp
+	call _RstFarCall.do_farcall
+	jmp FarCall_hl.return
```
Rangi42, line 15: These should be top-level labels since they're referenced externally

home/farcall.asm
```diff
@@ -2,131 +2,242 @@
...
+; +11 return address
+; +10 saved bank
+; +8  return path
+; +6  function pointer
+; +4  saved af
+; +2  saved de
+; +0  saved hl
+.do_farcall
+; same speed and size as add sp, -2 but preserves flags
+	dec sp
```
Rangi42, line 20: What if an interrupt happens here?

home/farcall.asm
```diff
@@ -2,131 +2,242 @@
...
+.do_farcall
+	push hl
+	push af
+	ldh a, [hROMBank]
+	ld hl, sp + 6
+	ld [hl], a
+	pop af
+	rst Bankswitch
+; same speed as ret but doesn't leave a stack address in hl
+; less likely to trash the stack if bugs happen
```
Rangi42, line 106: What's wrong with trashing the stack and crashing if something is this broken?

ariscop: I'd still expect it to crash but the keeping the return address from being trashed would make debugging easier

ariscop: Changed to use call _hl_

home.asm
```diff
@@ -63,15 +63,33 @@ INCLUDE "home/vba.asm"
 INCLUDE "home/vwf.asm"
 
 ; data used within home
+SECTION "Alternate music table", ROM0
```
Rangi42, line 66: Why? Technically a lot more of ROM0 could be single-table or few-routine sections, but would be clutter to do that.

home/farcall.asm
```diff
@@ -2,131 +2,233 @@
...
-; Updates the flags on the stack to their current value
-; without clobbering any other registers.
-	push af ; pushes current flags
+_RstFarCall::
+	call .do_farcall
+	jr _ReturnFarCall
+; Call the following dab pointer
+; preserves af, bc, de, hl
+
+; Stack layout
```
Rangi42, line 11: It's not clear at what point this layout is supposed to actually be the case during all the inc/dec sps and push/pops below.

home/farcall.asm
```diff
@@ -2,131 +2,233 @@
...
 	ret
 
-RetrieveAHLAndCallFunction:
-; Call the function at hl with restored values of a and hl.
+FarCall_de::
+; Call a:de.
+	scf
+	jr _DoFarCall
+
+; Stack layout:
```
Rangi42, line 122: It's not clear at what point this layout is supposed to actually be the case during all the inc/dec sps and push/pops below.

home/predef.asm
```diff
@@ -1,40 +1,70 @@
...
-	jr nz, .jump
-	push hl
-.jump
-	ld l, a
-	ldh a, [hROMBank]
+	dec sp
+	call .do_farcall
+	jmp _ReturnFarCall
+
+; Stack layout
```
Rangi42, line 8: I'd hate to have to modify the predef/farcall code again after this. :P

home.asm
ariscop: So the linker can sort them into gaps in home

Not attached to the change, just seemed like a good way to reduce work packing home with with minimal clutter

home/farcall.asm
ariscop: After the first set of dec sp+push, should i move the comment there?

home.asm
Rangi42: Yeah, please just leave this out, it's unrelated to the PR title.

home/farcall.asm
Rangi42: If that's the point where it's correct, yes please.

Rangi42 (commented):
Overall: Now that I've studied it the approach makes more sense, and I could edit/update it if needed. It does use a little more ROM0 space and a few more cycles per farcall than before, and the main advantage is farcalls being reentrant/safe in interrupts, plus four freed HRAM bytes. Up to you, @FredrIQ :)

PR #820, opened 2023-03-11
Rewrite argument handling in optimize.py

ariscop:
Allows passing an arbitrary set of files/folders

### DIFF
--- a/utils/optimize.py
+++ b/utils/optimize.py
@@ -5,8 +5,11 @@
 Search all .asm files for N code lines in a row that match some conditions.
 """
 
+from argparse import ArgumentParser
 from collections import namedtuple
 from glob import iglob
+from pathlib import Path
+from sys import stderr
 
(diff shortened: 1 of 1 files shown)


utils/optimize.py
```diff
@@ -642,8 +638,27 @@ def isNotReallyHram(code):
...
+total_count = 0
+for path in args.path:
+	if not path.exists():
+		print("File not found:", path)
+		next
+	if path.is_file():
+		total_count += check_file(path)
+		next
+	for f in path.rglob("*.asm"):
+		total_count += check_file(f)
```
Rangi42, line 659:
```suggestion
total_count = 0
for path in args.path:
	if not path.exists():
		print("File not found:", path, file=sys.stderr)
		continue
	if path.is_file():
		total_count += check_file(path)
	else:
		for f in path.rglob("*.asm"):
			total_count += check_file(f)
```

PR #821, opened 2023-03-11
List tools/lzcomp as a prereq for .lz files

ariscop:


Makefile
```diff
@@ -264,7 +264,7 @@ gfx/pokemon/%/frames.asm: gfx/pokemon/%/front.animated.tilemap gfx/pokemon/%/fro
 	$Qtools/pokemon_animation -f $^ > $@
 
 
-%.lz: %
+%.lz: % $Qtools/lzcomp
```
Rangi42, line 267:
```suggestion
%.lz: % tools/lzcomp
```

PR #823, opened 2023-03-12
Check for calls to rst functions in optimize.py

ariscop:


Rangi42:
Thanks, but optimize.py is meant to be usable for other projects. This could be its own script, like farcheck.py.

PR #824, opened 2023-03-12
Check for calls to rst functions in farcheck.py

ariscop:
Pretty sure this isn't what you meant by 'like farcheck.py' but it does have the advantage of no upkeep if rst symbols change


Rangi42:
Nice how this doesn't require hard-coding, thanks.

PR #827, opened 2023-03-15
Move hRTC* to wram

ariscop:
Ends up being pretty much equivalent, uses a few less rom0 bytes and 5 less hram bytes

Reduces the delta with the no-rtc build

UpdateTime already assumed rSVBK == 1 but only on no-RTC, so it's possible bugs in that build would be exposed to rtc-enabled builds


Rangi42:
Was this tested with rtc and no-rtc, including saving and reloading?

ariscop:
Yeah, including moving a save between rtc and no-rtc at least once

PR #828, opened 2023-03-17
Audio space saving

ariscop:
```
Before: ROMX: 2003678 bytes used / 11554 free in 123 banks
After:  ROMX: 2003384 bytes used / 11848 free in 123 banks
```


Rangi42:
Interesting, you've been comparing the regular builds; I'm used to looking at the output from `make debug` (was `77714/2097152 (3.71%)`, now ...).

PR #829, opened 2023-03-19
Compacting music commands

ariscop:
```
Before: ROMX: 2004054 bytes used / 11178 free in 123 banks
After:  ROMX: 2002147 bytes used / 13085 free in 123 banks
 +1907 bytes
```

Changes stereo panning and duty cycle to avoid an argument byte, and a few other things

Music source compatible, with the exception of `stereo_panning FALSE, FALSE` which is no longer valid

### DIFF
--- a/audio/engine.asm
+++ b/audio/engine.asm
@@ -1318,7 +1318,6 @@ MusicCommands:
 	dw Music_SlidePitchTo ; pitch wheel
 	dw Music_Vibrato ; vibrato
 	dw Music_ToggleNoise ; music noise sampling
-	dw Music_Panning ; force panning
 	dw Music_Volume ; volume
 	dw Music_Tone ; tone
 	dw Music_TempoRelative ; global tempo
@@ -1328,6 +1327,7 @@ MusicCommands:
 	dw Music_SFXPriorityOff ; sfx priority off
 	dw Music_StereoPanning ; stereo panning
 	dw Music_SFXToggleNoise ; sfx noise sampling
+	dw DoNothing ; $EF
 	dw DoNothing ; $F0
 	dw DoNothing ; $F1
 	dw DoNothing ; $F2
@@ -1796,11 +1796,6 @@ Music_StereoPanning:
 	bit STEREO, a
 	; skip param
 	jr z, GetMusicByte
-	; fallthrough
-
-Music_Panning:
-; force panning
-; params: 1
 	call SetLRTracks
 	call GetMusicByte
 	ld hl, wChannel1Tracks - wChannel1
--- a/audio/engine.asm
+++ b/audio/engine.asm
@@ -1429,7 +1429,7 @@ Music_LoopChannel:
 	bit SOUND_LOOPING, [hl] ; has the loop been initiated?
 	jr nz, .checkloop
 	and a ; loop counter 0 = infinite
-	jr z, .loop
+	jr z, Music_JumpChannel
 	; initiate loop
 	dec a
 	set SOUND_LOOPING, [hl] ; set loop flag
@@ -1443,7 +1443,6 @@ Music_LoopChannel:
 	and a ; are we done?
 	jr z, .endloop
 	dec [hl]
-.loop
 	jr Music_JumpChannel
 
 .endloop
--- a/audio/engine.asm
+++ b/audio/engine.asm
@@ -1326,9 +1326,9 @@ MusicCommands:
 	dw Music_NewSong ; new song
 	dw Music_SFXPriorityOn ; sfx priority on
 	dw Music_SFXPriorityOff ; sfx priority off
-	dw DoNothing ; $EE
 	dw Music_StereoPanning ; stereo panning
 	dw Music_SFXToggleNoise ; sfx noise sampling
+	dw DoNothing ; $F0
 	dw DoNothing ; $F1
 	dw DoNothing ; $F2
 	dw DoNothing ; $F3
--- a/audio/engine.asm
+++ b/audio/engine.asm
@@ -1326,7 +1326,7 @@ MusicCommands:
 	dw Music_SFXPriorityOn ; sfx priority on
 	dw Music_SFXPriorityOff ; sfx priority off
 	dw Music_StereoPanning ; stereo panning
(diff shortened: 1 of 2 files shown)


audio/engine.asm
```diff
@@ -1304,39 +1304,39 @@ MusicCommands:
...
-	dw MusicF2 ; nothing
-	dw MusicF3 ; nothing
-	dw MusicF4 ; nothing
-	dw MusicF5 ; nothing
-	dw MusicF6 ; nothing
-	dw MusicF7 ; nothing
-	dw MusicF8 ; nothing
+	dw Music_PanLeft ; stereo panning
+	dw Music_PanRight
+	dw Music_PanCentre
```
Rangi42, line 1330:
```suggestion
	dw Music_PanCenter
```

audio/engine.asm
```diff
@@ -1763,10 +1709,12 @@ Music_SoundStatus:
...
 
-Music_DutyCycle:
+Music_DutyCycle0:
+Music_DutyCycle1:
+Music_DutyCycle2:
+Music_DutyCycle3:
 ; duty cycle
-; params: 1
-	call GetMusicByte
+	ld a, [wCurMusicByte]
```
Rangi42, line 1717: If this depends on the numeric values of the music commands, there should be an `assert` to that effect.

audio/engine.asm
```diff
@@ -1826,21 +1774,27 @@ Music_ForceOctave:
...
-	jr z, GetMusicByte
+Music_PanLeft:
+	ld a, $F0
+	jr _stereo_pan
+
+Music_PanRight:
+	ld a, $0F
+	jr _stereo_pan
+
+Music_PanCentre:
```
Rangi42, line 1786:
```suggestion
Music_PanCenter:
```

audio/engine.asm
```diff
@@ -1826,21 +1774,27 @@ Music_ForceOctave:
...
+	ld a, $FF
 	; fallthrough
+_stereo_pan:
```
Rangi42, line 1789:
```suggestion
_ForcePanning:
```

audio/engine.asm
ariscop: Not sure where best to do that, `assert duty_cycle_cmd & 3 == 0` in the macro file maybe?
Its arguably enforced by the jump table, octave does much the same and lacks an assert. the exact value isn't important, just the two lowest bits

ariscop:
$F0 -> $F8 were left open in part for relative branch commands, unhappy with the implementation i have right now though

audio/engine.asm
Rangi42: Good point about `octave`, it should have one too (in pokecrystal as well). In the command body is fine, like where there's currently a `; set octave based on lo nybble of the command` comment. Since shifting the values by 1 would break the implementation.

ariscop: Added asserts for duty_cycle and octave, plus sound_ret while i was there

Added them to the macro file because the _cmd defines are otherwise never used outside there and break the pattern of defines being all upper case

audio/engine.asm
```diff
@@ -1826,21 +1774,27 @@ Music_ForceOctave:
...
-	jr z, GetMusicByte
+Music_PanLeft:
+	ld a, $F0
+	jr _ForcePanning
+
+Music_PanRight:
+	ld a, $0F
+	jr _ForcePanning
+
+Music_Pancenter:
```
Rangi42, line 1786:
```suggestion
Music_PanCenter:
```

macros/scripts/audio.asm
```diff
@@ -48,15 +48,25 @@ ENDM
 	const_def $d0
 DEF FIRST_MUSIC_CMD EQU const_value
 
-	const octave_cmd ; $d0
+	const octave_cmd
 MACRO octave
 	assert 1 <= (\1) && (\1) <= 8, "octave must be 1-8"
 	db octave_cmd + 8 - (\1) ; octave
 ENDM
+assert octave_cmd & $7 == 0, "octave_cmd must be 3 bit aligned"
```
Rangi42, line 56:
```suggestion
	const octave_cmd
	assert octave_cmd & %111 == 0, "octave_cmd must be 3-bit aligned"
MACRO octave
	assert 1 <= (\1) && (\1) <= 8, "octave must be 1-8"
	db octave_cmd + 8 - (\1) ; octave
ENDM
```

macros/scripts/audio.asm
```diff
@@ -48,15 +48,25 @@ ENDM
...
 
 	const_skip 7 ; all octave values
 
-	const note_type_cmd ; $d8
+	const duty_cycle_cmd
+MACRO duty_cycle
+	assert 0 <= (\1) && (\1) <= 6, "duty cycle must be 0-6"
+	db duty_cycle_cmd | (\1 & 3) ; values 4-6 fold into 0-2
+ENDM
+assert duty_cycle_cmd & $3 == 0, "duty_cycle_cmd must be 2 bit aligned"
```
Rangi42, line 65:
```suggestion
	const duty_cycle_cmd
	assert duty_cycle_cmd & %11 == 0, "duty_cycle_cmd must be 2-bit aligned"
MACRO duty_cycle
	assert 0 <= (\1) && (\1) <= 6, "duty cycle must be 0-6"
	db duty_cycle_cmd | (\1 & 3) ; values 4-6 fold into 0-2
ENDM
```

macros/scripts/audio.asm
```diff
@@ -311,13 +253,14 @@ MACRO sound_loop
...
 	db sound_call_cmd
 	dw \1 ; address
 ENDM
 
-	const sound_ret_cmd ; $ff
+	const sound_ret_cmd
 MACRO sound_ret
 	db sound_ret_cmd
 ENDM
+assert sound_ret_cmd == $ff, "sound_ret must be $ff"
```
Rangi42, line 266:
```suggestion
	const sound_ret_cmd
	assert sound_ret_cmd == $ff, "sound_ret_cmd must be $ff"
MACRO sound_ret
	db sound_ret_cmd
ENDM
```

PR #831, opened 2023-03-22
Sound engine optimization

ariscop:
Removes dead code, unused variables, and elides many hl loads

Commands like new_song and restart_channel are currently unused and were not tested as a result. it may make sense to remove them

238 fewer rom bytes and 116 less wram0

### DIFF
--- a/audio/engine.asm
+++ b/audio/engine.asm
@@ -1570,25 +1570,23 @@ Music_ToggleNoise:
 	ld hl, wChannel1Flags - wChannel1
 	add hl, bc
 	bit SOUND_NOISE, [hl]
-	jr z, .on
 	; turn noise sampling off
 	res SOUND_NOISE, [hl]
-	ret
-
-.on
-	; turn noise sampling on
+	ret nz
+	; was off, turn noise sampling on instead
 	set SOUND_NOISE, [hl]
-	ld a, [wCurChannel]
-	cp CHAN5
-	jr c, Music_ChangeNoiseSampleSet
-; Channel 8 uses sfx sample set
-	call GetMusicByte
-	ld [wSFXNoiseSampleSet], a
-	ret
+; fallthrough
 
 Music_ChangeNoiseSampleSet:
+	ld a, [wCurChannel]
+	cp CHAN5
+	ld hl, wMusicNoiseSampleSet
+	jr c, .music
+; Channel 8 uses sfx sample set
+	inc hl ; wSFXNoiseSampleSet
+.music
 	call GetMusicByte
-	ld [wMusicNoiseSampleSet], a
+	ld [hl], a
 	ret
 
 Music_NoteType:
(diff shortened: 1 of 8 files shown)


home/audio.asm
```diff
@@ -252,18 +273,8 @@ WaitSFX::
...
-	bit 0, [hl]
-	jr nz, .wait
-	ld hl, wChannel7Flags
-	bit 0, [hl]
-	jr nz, .wait
-	ld hl, wChannel8Flags
-	bit 0, [hl]
-	jr nz, .wait
+	call CheckSFX
+	jr c, .wait
```
Rangi42, line 277: Looks like the return code of `CheckSFX` can just be simplified to `z/nz` and have `bit SOUND_CHANNEL_ON, a / ret`.

Rangi42:
> Commands like new_song and restart_channel are currently unused and were not tested as a result. it may make sense to remove them

That's alright; did you test the Sw/Sh Gym leader battle music though? Try fighting Piers in the Goldenrod Underground and note how the the track changes when your mon faints, when his mon does, and when it's the last mon.

audio/engine.asm
```diff
@@ -803,17 +760,14 @@ HandleTrackVibrato:
 	and a
 	jr nz, .subexit
 	; is the extent nonzero?
-	ld hl, wChannel1VibratoExtent - wChannel1
-	add hl, bc
-	ld a, [hl]
+	inc hl
```
Rangi42, line 763: How about an `assert VIBRATO_DELAY_COUNT + 1 == VIBRATO_EXTENT`? Likewise for other implicit "adjust hl to point to this field" optimizations. (In case you do more and they shift.)

ariscop:
Didn't know where to find sw/sh song in game until now, it seems fine; ultimately i only changed comments in-case they are ever uncommented. found them while looking for songs that use the pitch sweep command

Been meaning to look into those conditional branches and maybe add support to the music player

Rangi42:
> Didn't know where to find sw/sh song in game until now, it seems fine; ultimately i only changed comments in-case they are ever uncommented. found them while looking for songs that use the pitch sweep command
> 
> Been meaning to look into those conditional branches and maybe add support to the music player

Okay, thanks; I wanted to be sure because you removed the `wChannel*JumpCondition`s, but I guess those aren't actually used by `sound_jump_if`.

home/audio.asm
ariscop: The call sites already expect c but there aren't many, i'll change it

ariscop:
The JumpConditions are from MusicEE in pokecrystal, which was removed from polished some time ago

Rangi42:
Sorry about the merge conflicts :P

audio/engine.asm
Rangi42: I'll merge for now as-is, but more asserts here (and maybe in move_mon.asm, IIRC it also does hli/hld assumptions) would be nice.

ariscop: Left comments on any implicit reference that crosses a field so i can grep for usage

Happens kind of randomly throughout the file though, asserting each instance would add noise from redundant asserts

PR #832, opened 2023-03-22
Move music player ram to Misc 480

ariscop:
Memory used by the audio engine moved into its section

As a side effect, settings are preserved when re-entering the
music player

+2 trivial patches

### DIFF
--- a/audio/music_player.asm
+++ b/audio/music_player.asm
@@ -171,18 +171,7 @@ RenderMusicPlayer:
 	xor a
 	ldh [hOAMUpdate], a ; we will manually do it in LCD interrupt
 
-	ld hl, wChannelSelectorSwitches
-.ch_label_loop:
-	ld [wChannelSelector], a
-	ld a, [hli]
-	push hl
-	call DrawChannelLabel
-	pop hl
-	ld a, [wChannelSelector]
-	inc a
-	cp NUM_MUSIC_CHANS
-	jr nz, .ch_label_loop
-
+	call RedrawChannelLabels
 	call DelayFrame
 
 	ldh a, [rSVBK]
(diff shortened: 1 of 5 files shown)


ariscop:
misc 480 isn't in use during the main menu so it can safely be used for the music player. its state doesn't need to persist after the player exits, so dedicated wram0 isn't needed

Variables accessed by the audio engine (noise hit, tempo adjust, etc) however do need to persist and were moved to the audio ram section as a result

ariscop:
Actually don't see a reason it couldn't be made available during gameplay, as an item or map event

Misc 480 is were slots, card flip, pokedex, etc are allocated

audio/music_player.asm
```diff
@@ -172,16 +172,15 @@ RenderMusicPlayer:
...
 .ch_label_loop:
 	ld [wChannelSelector], a
 	ld a, [hli]
 	push hl
 	call DrawChannelLabel
 	pop hl
 	ld a, [wChannelSelector]
-	dec a
-	cp -1
+	inc a
```
Rangi42, line 182: This makes the `wChannelSelector` iterate ascending not descending, but the values read from `wChannelSelectorSwitches` are still ascending. Is that OK?

home/lcd.asm
```diff
@@ -35,13 +35,6 @@ LCDMusicPlayer::
 	ld [oamSprite01YCoord], a
 	ld [oamSprite02YCoord], a
 
-	ldh a, [hMPState]
```
Rangi42, line 38: (This was a leftover from https://github.com/Rangi42/polishedcrystal/commit/7206aecb7021adbe412ecb3719136d7da76b2c45.)

ram/wram0.asm
```diff
@@ -258,12 +265,6 @@ wSelectorCur:: db
 ; song editor
 wChannelSelector:: db
 wAdjustingTempo:: db
-; audio engine input
-wChannelSelectorSwitches:: ds 4
```
Rangi42, line 262: Now that this is no longer in the range of `wMusicPlayerWRAMEnd - wMusicPlayerWRAM`, does the "Initialize wMusicPlayerWRAM" need updating?

audio/music_player.asm
ariscop: This was a bug, both are in ascending order, but `wChannelSelectorSwitches` was cleared entry so it always read 0

ram/wram0.asm
ariscop: This does mean no easy way To clear them, will make b from the editor do so

Rangi42: Thanks!

Rangi42:
> Actually don't see a reason it couldn't be made available during gameplay, as an item or map event

Key item: Music Player
image

PR #833, opened 2023-03-23
Check rom hash for save state detection, instead of save version

ariscop:
A separate check already exists on save load, both prevents bypassing that check and loading save states between rom revisions

Moved wRomChecksum to a fixed location, part of the options section


home/header.asm
```diff
@@ -170,4 +170,6 @@ if DEF(ANALOGUE_POCKET)
 endc
 
 	; The rest of the header is handled by rgbfix.
-	ds $0150 - @, $00
+	ds $014e - @, $00
+
+RomChecksum:: ds 2
```
Rangi42, line 175:
```suggestion
RomHeaderChecksum::
	ds 2, $00
```

Rangi42:
What's the goal here? There are many trivial commits which would change the ROM checksum but shouldn't prevent loading saves or emulator save states.

home/header.asm
ariscop: There's a separate checksum for the header alone at $14d, while $14e-$14f covers all banks

ariscop:
This doesn't break save loading on rom revision, just save states, which most of the time would work but lots of pointers float around at link time, notably map data, and can move between revisions if any section changes size

ram/wram0.asm
```diff
@@ -1392,3 +1392,12 @@ wDaysSince:: db
...
 wOptionsBuffer:: db
+
+	ds 1 ; unused
+
+; Contains a copy of the rom checksum, read from the header. Used as
+; protection against people trying to load a save state for a save in
+; a different rom version.
+wRomChecksum:: dw
+
+assert @ == WRAM0_End
```
Rangi42, line 1403: `wRomChecksum` could have its own section, `org`ed in the linkerscript.

ariscop: put it there simply because there were 3 bytes of slack from options already having an org. Would a separate section be preferable?

Rangi42: Yes please, instead of an `assert`. You can redundantly do `ROMX[$dffe]` too, same as how the header sections are fixed.

engine/init.asm
```diff
@@ -63,17 +63,11 @@ _Init::
...
-	ld a, BANK(wGameVersion)
-	ldh [rSVBK], a
-	ld a, HIGH(SAVE_VERSION)
-	ld [wGameVersion], a
-	ld a, LOW(SAVE_VERSION)
-	ld [wGameVersion + 1], a
-	pop af
-	ldh [rSVBK], a
+	; Write checksum to WRAM for save state check
+	ld a, [RomChecksum]
```
Rangi42, line 67: `RomHeaderChecksum` would be more clear where it's coming from.

Rangi42: (Also keeping it distinct from `wRomChecksum` so it doesn't look like a typo; it's unusual to see a load not from Hungarian-notated RAM labels.)

ariscop:
For reference there is a save version check, `engine/menus/save.asm: VerifyGameVersion`

This generalizes the runtime check for loading save states between builds, which is not guaranteed to work while regular save/load is

ariscop:
This should probably have a distinct crash message like VBABSODMessage

Rangi42:
Doesn't replacing `ERR_VERSION_MISMATCH` with `ERR_CHECKSUM_MISMATCH` mean you can't load an older save file?

ariscop:
No, ERR_VERSION_MISMATCH never actually checked the save file, it wrote the save version to ram and checked that
Changing the hash at $14e-$14f after boot is the only thing that triggers ERR_CHECKSUM_MISMATCH

FredrIQ:
> Doesn't replacing `ERR_VERSION_MISMATCH` with `ERR_CHECKSUM_MISMATCH` mean you can't load an older save file?

To be clear, there were 2 save version checks.
There was the regular check, VerifyGameVersion, handled upon loading a save. This prints the "game and save version doesn't match, if the save version is too old, consult documentation for how to update" (etc) message. Therre was also a separate handler in VBlank that compared the game version with the version stored in WRAM. These could be mismatched if the user tried to bypass VerifyGameVersion by loading a save state. I added this check after seeing people on Prism's discord trying this veriy thing with its own version checker.

A ROM checksum verification would catch all these cases (making them redundant), and more on top. Hence why there's no point in a version check watchdog if we added a rom checksum one.

itsdarsh:
I am going to modify this so that it doesn't bsod on debug versions (otherwise debugging will become a pain in the ass for me) but this seems very sensible otherwise.

PR #834, opened 2023-03-24
Optimize CountSetBits

ariscop:
CountTMsHMs from 1040 cycles to 457, with full set of tms

Since the performance of CountSetBits is largely unimportant, PopCountA is not inlined so it can be reused elsewhere


itsdarsh:
Keeping this in mind in case we meaningfully expand TMs or dex.

PR #835, opened 2023-03-26
Fix Violet gym guy throwing you across the street

ariscop:
Likely introduced by 9b322012

This uses the warp event instead of hard coding warp coordinates


Rangi42:
Interesting. If you don't want to just fix the coordinates, want to update other such scripts which cause the player to exit/enter a building? E.g. from Route 34 to the Daycare, or from Ecruteak Gym to Ecruteak City.

ariscop:
Did have a quick look, i'm not sure how many cases of scripted use of a warp carpet there are but most cases of warp could be warpmod which specifies a spawn number instead of coordinates

will have a go at it

ariscop:
It takes a bit to trigger each event to test them, happy to look into this more but preferably in a different pr

PR #836, opened 2023-03-27
Wram housekeeping

ariscop:
Mostly removes unused constraints in layout.link

Misc 480/1326 are both 4 bit aligned, nothing seems to depend on that fact anymore (wMisc likely used to) but it cost nothing to preserve that

The only alignment sensitive code i found dealt with wBGMapBuffer, for which asserts are provided

### DIFF
--- a/home/video.asm
+++ b/home/video.asm
@@ -70,6 +70,11 @@ UpdateBGMapBuffer::
 	ld bc, wBGMapPalBuffer
 	ld de, wBGMapBuffer
 
+; We increment the low byte of a pointer, so ensure these buffers
+; dont cross a 256 byte boundary
+assert HIGH(wBGMapBuffer) == HIGH(wBGMapBuffer + 47)
+assert HIGH(wBGMapPalBuffer) == HIGH(wBGMapPalBuffer + 47)
+
 .next
 ; Copy a pair of 16x8 blocks (one 16x16 block)
 
--- a/home/video.asm
+++ b/home/video.asm
@@ -72,8 +72,8 @@ UpdateBGMapBuffer::
 
 ; We increment the low byte of a pointer, so ensure these buffers
 ; dont cross a 256 byte boundary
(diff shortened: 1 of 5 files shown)


home/video.asm
```diff
@@ -70,6 +70,11 @@ UpdateBGMapBuffer::
 	ld bc, wBGMapPalBuffer
 	ld de, wBGMapBuffer
 
+; We increment the low byte of a pointer, so ensure these buffers
+; dont cross a 256 byte boundary
+assert HIGH(wBGMapBuffer) == HIGH(wBGMapBuffer + 47)
```
Rangi42, line 75: The 47s are suspicious. Should those be `wBGMapBufferEnd` and `wBGMapPalBufferEnd`?

layout.link
```diff
@@ -220,22 +220,10 @@ WRAM0
 	org $c000
 	"Stack"
 	align 8
-	"Audio RAM"
-	"WRAM 0"
-	org $c308
```
Rangi42, line 225: Don't these exist to minimize WRAM shifting, which breaks saves and needs darsh to update the save patch?

ariscop: Will have to look into that, was not aware anything from either of those sections were saved

ariscop: wOptions, wPlayerData, wPokemonData, wCurMapData, seem to be what gets saved and all of those besides wOptions reside in wramx

My debug save seems to load fine

home/video.asm
ariscop: Yeah i should just add labels, would also prevent it from breaking if they change size

layout.link
ariscop: org directives have been there since linker scripts were first added in d9f530f8. can't find any explicit reasoning

`org $c308` is from 91f0530e

itsdarsh: To close this point, no, this doesn't affect saves. As ariscop said, everything I care about is in wramx.

PR #847, opened 2023-05-18
fixed bike and surf music in overcast areas

VermilionCosmos:
Removed the handling for overcast special music in order to restore bike and surf music in overcast areas.


VermilionCosmos:
Accidentally attached to wrong branch.

PR #848, opened 2023-05-18
Fix for bicycle and surf music in overcast areas.

VermilionCosmos:
Removed overcast music handling to restore bike and surf music in overcast areas.

### DIFF
--- a/home/audio.asm
+++ b/home/audio.asm
@@ -70,26 +70,6 @@ _LoadMusicWord::
 
 CheckSpecialMapMusic:
 ; Returns z if the current map has a special music handler.
-
-	ldh a, [hROMBank]
-	push af
-	ld a, BANK(GetOvercastIndex)
-	rst Bankswitch
-	call GetOvercastIndex ; far-ok
-	ld b, a
-	pop af
-	rst Bankswitch
-	ld a, b
-	and a
-	jr z, .check_special
-	ld hl, .OvercastMusicHandler
-	xor a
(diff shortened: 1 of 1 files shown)


vulcandth:
Instead of deleting the overcast music handler code.. we should instead add a check for PlayerState to determine if the player is actively riding a bike or not.

vulcandth:
Something like:

```asm
ld a, [wPlayerState]
and a
jr z, .check_special

```

`and a` checks if the register a is zero.. aka `PLAYER_NORMAL` (not surfing or biking).

itsdarsh:
I'd like to know why it was added in the first place before merging this. There appears to be some intentionality behind the code as written.

VermilionCosmos:
So the I am unsure if the player state needs to be checked because there is already functionality that just disables bike and surf music. The only thing that the Overcast Handler does is see if its Overcast, then run the GetMapMusic command which is only used for the purpose of turning off bike and surf music. I think its more that the Overcast Handler doesn't seem like it has a purpose right now, but maybe was meant to have one and left there, which ended up just removing functionality.

(I could be completely wrong, but this is my understanding.)

VermilionCosmos:
To clarify, there is already a thing called "GetMapMusic" which only purpose is to disable surf and bike music. Overcast handler just seemed to be calling it, but had no special music of its own.

Rangi42:
Yes, this code was intentional.

PR #850, opened 2023-05-20
Throat Spray now works for both damaging and non-damaging moves.

VermilionCosmos:
Created a call just for Throat Spray to account for both damaging and non damaging sound moves. Uses wAttackMissed to see if the attack hit.

### DIFF
--- a/engine/battle/effect_commands.asm
+++ b/engine/battle/effect_commands.asm
@@ -3253,24 +3253,22 @@ CheckThroatSpray:
 	ld a, [wBattleEnded]
 	and a
 	ret nz
+
 	ld a, [wAttackMissed]
 	and a
-	jr z, .do_it
-	ret
+	ret nz
 	
-.do_it
 	call HasUserFainted
 	ret z
+
 	predef GetUserItemAfterUnnerve
 	push bc
 	call GetCurItemName
 	pop bc
 	ld a, b
 	cp HELD_THROAT_SPRAY
-	jr z, .item_valid
-	ret
+	ret nz
 	
-.item_valid
 	push bc
 	ld a, BATTLE_VARS_MOVE_ANIM
 	call GetBattleVar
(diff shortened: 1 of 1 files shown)


VermilionCosmos:
As a note, Soundproof needs further investigation. It doesn't activate for non-damaging sound moves, and when it does, it will need a special case for Heal Bell.

FredrIQ:
Soundproof shouldn't prevent Heal Bell from working. This was a thing in early generations, but not modern ones.

itsdarsh:
> As a note, Soundproof needs further investigation. It doesn't activate for non-damaging sound moves, and when it does, it will need a special case for Heal Bell.

I may be misinterpreting this but this isn't entirely correct, Soundproof does (or at least should) block all sound moves from the opponent, damaging or not. It never blocks the *user's* own sound-based moves, such as Perish Song and Heal Bell. No special case exception for Heal Bell is needed, rather it shouldn't even check the user's own moves in the first place.

itsdarsh:
I should just note for the record that despite all the quibbles I've raised, this is very impressive work for someone's first time tweaking the battle engine. It's not easy to grok!

itsdarsh (changes requested):

engine/battle/effect_commands.asm
```diff
@@ -3248,6 +3249,37 @@ CheckStatHerbs:
...
+	ld a, [wAttackMissed]
+	and a
+	ret nz
+	
+	call HasUserFainted
+	ret z
+
+	predef GetUserItemAfterUnnerve
+	push bc
+	call GetCurItemName
```
itsdarsh, line 3266: Move this to above line 3273 and test if everything still works as expected. It may save an extra push/pop.

itsdarsh (changes requested):

engine/battle/effect_commands.asm
VermilionCosmos: Works!

itsdarsh (approved):

PR #862, opened 2023-07-01
Audio fixes

dannye:
The audio macro upgrade script put too many `drum_note` commands in swsh/gymleaderbattle.asm due to the channel definitions being interwoven in that file. The first commit in this PR fixes those and doesn't change the rom checksum.

The second commit makes a few fixes that allow those files to be opened in Crystal Tracker.
* Several songs used `duty_cycle` in channel 3 but that seems to be a mistake from a conversion tool. The duty cycle value appears to be the intended channel 3 wave id and usually matches the nearest wave id from a nearby `volume_envelope` or `note_type` command. These `duty_cycle` commands certainly do nothing unless these music branches are also used by channels 1 or 2 -- which I don't believe to be the case -- or if audio/engine.asm has been modified such that `duty_cycle` actually has a meaning in channel 3 -- which I also don't believe to be the case. The duty cycle values that don't match a nearby wave id are probably leftovers from the original editing process and won't be missed.
* `tempo` commands should only be used by the first channel in the song. prism/trainerbattle.asm used a tempo command at the start of channel 2 (misleadingly named channel 1). Technically harmless (**only** because the two tempo values are the same) but best to eliminate it.
* xy/titlescreen.asm used a couple local defs for some drum ids. There's technically nothing wrong with this, so feel free to veto this change, but the constants aren't necessary and prevent the file from opening in Crystal Tracker. Especially given that there were only 2 constants *and* they were equal to the same value, not much advantage to using the constants unless you wanted the ability to quickly separate the notes into two different drums by changing just one of the constants.

In addition to that, I noticed there are 5 songs that have a major desync between their channels after the main body loop:

> bw/celestialtower.asm
dppt/route205.asm
rse/battlefactory.asm
xy/gymleaderbattle.asm
xy/titlescreen.asm

You can see the desync by opening them in Crystal Tracker, waiting for the lag, and seeing how wide the timeline is as a result of extending the song until all channels arrive at the beginning at the same time again.


audio/music/xy/gymleaderbattle.asm
```diff
@@ -2853,102 +2853,6 @@ Music_GymLeaderBattleXY_Ch4_loop:
 	drum_note 7, 2
 	drum_note 12, 1
 	drum_note 4, 1
-	drum_note 7, 2
```
dannye, line 2856: This repetitive pattern just went on for 8 measures longer than it should have. This is the only debatably subjective fix imo.

audio/music/rse/battlefactory.asm
```diff
@@ -17,11 +17,12 @@ Music_BattleFactoryRSE_Ch1:
 	note_type 12, 4, 7
 	rest 12
 	rest 16
+	rest 8
 Music_BattleFactoryRSE_Ch1_loop:
 	duty_cycle 2
 	vibrato 18, 2, 4
-	rest 16
 	note_type 6, 4, 7
+	rest 16
```
dannye, line 25: This wasn't actually a desync, but this first rest of the main loop played at speed 12 on the first iteration and then played at speed 6 in all following iterations, which should produce a warning in Crystal Tracker but doesn't yet. The much less convoluted thing to do is to put the appropriate speed 12 rest before the loop and the appropriate speed 6 rest at the start of the loop. The result in-game is the same though.

PR #865, opened 2023-07-07
Fix for FastShip CaptainsCabin scene transition

vulcandth:
This is still not quite perfect yet... but its better. The main issue here has to do with the fact that this scene has the player move through the void from one room to another while the pals a "faded out". This triggers the dynpal system when new objects are loaded. 

Also it seems that `Special_FadeBlackQuickly` was setting the BG pals to grey.. and all OBJ pals to grey except for OBJ0 which was being set to Black. Since the player is typically OBJ0... They show up as black on a grey background in this scene.... 

So I made a `FadeToBlackQuickly` for now... Which just fades ALL pals to black.. and then we reload them before fading back in.

The fade in is still a little gittery.. i'll look into this a bit more. I'll also take suggestions.

### DIFF
--- a/home/palettes.asm
+++ b/home/palettes.asm
@@ -257,3 +257,18 @@ Special_ReloadSpritesNoPalettes::
 	ld a, 1
 	ldh [hCGBPalUpdate], a
 	jmp DelayFrame
+
+SetBlackObjectPals::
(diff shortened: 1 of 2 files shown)


vulcandth:
Smooth as butter now.

PR #869, opened 2023-08-03
Fix `*hl++|*hl-- = b|c|d|e` "Good" comment

vulcandth:


Rangi42:
I need to update the wiki too.

https://github.com/pret/pokecrystal/wiki/Optimizing-assembly-code#load-a-constant-into-hl-and-increment-or-decrement-hl

This should be "Load a value into `[hl]` and increment or decrement `hl`, and should document this case of `ld [hl], b|c|d|e / inc|dec hl` (2 bytes, 4 cycles) to `ld a, b|c|d|e / ld [hli|hld], a` (2 bytes, 3 cycles).

Rangi42:
Documented: https://github.com/pret/pokecrystal/wiki/Optimizing-assembly-code#load-a-register-into-hl-and-increment-or-decrement-hl

utils/optimize.py
```diff
@@ -338,7 +338,7 @@ def isNotReallyHram(code):
 ],
 '*hl++|*hl-- = b|c|d|e': [
 	# Bad: ld [hl], b|c|d|e / inc|dec hl (unless you can't use a)
```
Rangi42, line 340:
```suggestion
	# Bad: ld [hl], b|c|d|e|h|l / inc|dec hl (unless you can't use a)
```

utils/optimize.py
```diff
@@ -338,7 +338,7 @@ def isNotReallyHram(code):
...
-	# Good: ld [hli|hld], a / ld b|c|d|e, a
+	# Good: ld a, b|c|d|e / ld [hli|hld], a
```
Rangi42, line 341:
```suggestion
	# Good: ld a, b|c|d|e|h|l / ld [hli|hld], a
```

utils/optimize.py
```diff
@@ -338,7 +338,7 @@ def isNotReallyHram(code):
...
 	(lambda line1, prev: re.match(r'ld \[hl\], [bcde]', line1.code)),
```
Rangi42, line 342:
```suggestion
	(lambda line1, prev: re.match(r'ld \[hl\], [bcdehl]', line1.code)),
```

PR #880, opened 2023-08-16
Update evos_attacks.asm for proper Eevee location evolutions

TAG1Kilo:
In line with other fixes to the order of evolutions listed, this change lists the location evolutions for Glaceon and Leafeon before (instead of after) the happiness evolution conditions for Espeon and Umbreon. As the code stands, a sufficiently happy Eevee will not evolve into Glaceon or Leafeon, and rather attempt to become Espeon or Umbreaon, even if the player is in the correct location.


FredrIQ:
Thanks!

PR #891, opened 2023-10-13
Goldenrod Museum fix

mauvesea:
Fixes Goldenrod Museum pictures crashing the games. I've used the dba macro instead of fardw in the pic list. I'm not knowledgeable enough to understand why the fardw macro was having a hard issue loading the pic correctly, as that's what's leading to crashes. It only saves 7 bytes, so I'm not sure it's worth it.


SeasickShore:
I've also changed the pink NPC pal in overcast maps. It was using the fully bright day pal, and that's why it looked off.

data/events/paintings/pic_pointers.asm
```diff
@@ -1,11 +1,11 @@
...
-	fardw OmanytePuzzlePic
-	fardw AerodactylPuzzlePic
-	fardw HoOhPuzzlePic
+	dba HoOhPaintingPic
+	dba LugiaPaintingPic
+	dba BellTowerPaintingPic
+	dba KabutoPuzzlePic
+	dba OmanytePuzzlePic
+	dba AerodactylPuzzlePic
+	dba HoOhPuzzlePic
```
Rangi42, line 10:
```suggestion
	fardw HoOhPaintingPic
	fardw LugiaPaintingPic
	fardw BellTowerPaintingPic
	fardw KabutoPuzzlePic
	fardw OmanytePuzzlePic
	fardw AerodactylPuzzlePic
	fardw HoOhPuzzlePic
```

engine/gfx/load_pics.asm
```diff
@@ -276,7 +276,7 @@ GetPaintingPic:
 	ldh [hBGMapMode], a
 	ld hl, PaintingPicPointers
 	ld a, [wTrainerClass]
-	ld bc, 2
+	ld bc, 3
```
Rangi42, line 279:
```suggestion
	ld bc, 2
```

Rangi42:
Thanks for opening this! To fix the Goldenrod Museum paintings I think you just need to remove `inc hl` in `GetPaintingPic` (engine/gfx/load_pics.asm).

SeasickShore:
You're correct, just removing inc hl does the trick as well.

PR #895, opened 2023-11-19
Only enable LCD_STAT when needed.

vulcandth:
The idea is to only enable LCD_STAT when its actually needed. This can be found by enabling it whenever `hLCDCPointer` has a non-zero value... and disabling it when it is set back to zero.


vulcandth:
Okay reviewed everything thoroughly. This should be good to merge after your review.

PR #896, opened 2023-11-20
Fix lagging tail, quick claw, and quick draw

emilyploszaj:
Lagging tail was missing an `inc d` like `.set_priority` has to properly do the +1/-1 based on battle turn, instead doing +0/-2, which meant lagging tail did not work for enemy trainers. Further, a player with lagging tail and an enemy with quick claw could create -3, which is not considered part of the valid range for `d` in this routine. `.set_priority` was changed to be used for both increasing and decreasing priority called from the nicely named `.go_first` and `.go_last` labels.

I initially discovered this when implementing the Stall ability, the `.go_first` `.go_last` labels make that slightly more reasonable, and it makes the code here more readable, so I decided to bundle it with the lagging tail fix.

### DIFF
--- a/home/battle.asm
+++ b/home/battle.asm
@@ -762,7 +762,7 @@ CheckMoveSpeed::
 .quick_claw
 	ld a, 100
 	call BattleRandomRange
-	cp 20
+	cp c
 	pop de
 	ret nc
 .activate_item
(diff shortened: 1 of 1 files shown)


emilyploszaj:
I've also noticed that quick draw does not use battle random and quick claw compared against the c register which had... arbitrary content based on the random call, so I fixed those bugs as well.

PR #901, opened 2023-11-25
No RTC as an Init Option

vulcandth:
This implements no-rtc as an Initial in-game Option instead of a build option... Currently this replaces the Nuzlocke Mode Initial Game option.


katedering:
that'd be useful, but the nuzlockers would cry foul :P

so yeah, definitely implement this!

FredrIQ:
Before removal of Nuzlocke option, first encounter info needs to be conveyed in some way.

PR #918, opened 2023-12-21
Implement Mind's Eye as an actual ability.

katedering:
Should work exactly like in the official games. Also fixed (?) some bugs (?) in the original code, not sure why they were like that but when I changed it, the abilities started working properly.

### DIFF
--- a/engine/battle/effect_commands.asm
+++ b/engine/battle/effect_commands.asm
@@ -1590,10 +1590,10 @@ _CheckTypeMatchup:
 	jr nz, .end
 	call GetTrueUserAbility
 	cp SCRAPPY
-	jr nz, .end
+	jr z, .end
 	call GetTrueUserAbility
 	cp MINDS_EYE
-	jr nz, .end
+	jr nz, .TypesLoop
 .end
 	pop hl
 	ret
@@ -1853,7 +1853,8 @@ BattleCommand_checkhit:
 	jr nz, .avoid_evasion_boost
 	call GetTrueUserAbility
 	cp KEEN_EYE
-	jr nz, .check_opponent_unaware
+	jr z, .avoid_evasion_boost
+	call GetTrueUserAbility
 	cp MINDS_EYE
 	jr nz, .check_opponent_unaware
 .avoid_evasion_boost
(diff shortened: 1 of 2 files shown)


engine/battle/stats.asm
```diff
@@ -106,7 +106,10 @@ FarChangeStat:
 	jr z, .ability_check
 	cp KEEN_EYE
 	ld c, ACCURACY
-	jr nz, .check_item
+	jr z, .ability_check
+	cp MINDS_EYE
+	ld c, ACCURACY
+	jr z, .ability_check
```
itsdarsh, line 112: This should be `jr nz, .check_item`

engine/battle/effect_commands.asm
```diff
@@ -1851,6 +1854,8 @@ BattleCommand_checkhit:
 	call GetTrueUserAbility
 	cp KEEN_EYE
 	jr nz, .check_opponent_unaware
```
itsdarsh, line 1856: This should be `jr z, .avoid_evasion_boost`

engine/battle/effect_commands.asm
```diff
@@ -1590,7 +1590,10 @@ _CheckTypeMatchup:
 	jr nz, .end
 	call GetTrueUserAbility
 	cp SCRAPPY
-	jr nz, .TypesLoop
+	jr nz, .end
```
itsdarsh, line 1593: should be `jr z, .end`

engine/battle/effect_commands.asm
```diff
@@ -1590,7 +1590,10 @@ _CheckTypeMatchup:
...
+	call GetTrueUserAbility
+	cp MINDS_EYE
+	jr nz, .end
```
itsdarsh, line 1596: should be `jr nz, .TypesLoop`

itsdarsh (changes requested):
Please test the following changes

engine/battle/stats.asm
```diff
@@ -106,7 +106,10 @@ FarChangeStat:
 	jr z, .ability_check
 	cp KEEN_EYE
 	ld c, ACCURACY
-	jr nz, .check_item
+	jr z, .ability_check
+	cp MINDS_EYE
+	ld c, ACCURACY
```
itsdarsh, line 111: do not need to load `ACCURACY` into `c` again (you may want a comment to clarify that you still expect `c` to be `ACCURACY` here anyway even if it's not explicitly coded)

itsdarsh (approved):

PR #919, opened 2023-12-21
Orange Islands Overhaul

katedering:
This aims to properly cement Orange Islands as the 'de facto' endgame area, by both boosting the badge-base-level curve to cap at 68 instead of 61, and by drastically boosting the level of the wild Pokemon in Orange Islands. Additionally this adds the badge system to every trainer there and adds parties to the TODO bulbasaur trainers.


Rangi42 (changes requested):
Thanks, but these look unusually high. From the other `LEVEL_FROM_BADGES` uses in trainer parties, even `+ 10` is on the high end for non-boss trainers (and lower for wild mons, even with `-`s).

katedering:
I could drop the wild levels a bit but it never made sense to me that an area that you unlock post-e4 had wild encounters in the low-mid 30s, where most of Kanto has it in the 50s and up.

katedering:
image
this is the average wild level after ten badges, about the same as kanto's wild range in general

katedering:
orange

This is the first trainer in Orange Islands, encountered as soon as you reach Kanto. For reference, her team in base 9bit is two Level 55 Bellossom. The level range is exactly the same as default Polished 9bit IF you go there immediately after arriving in Kanto, even with the 'unusually high' level range, so I still think this is fine to push as-is.

data/trainers/parties.asm
```diff
@@ -8778,35 +8780,38 @@ AromaLadyGroup:
...
 	db -1 ; end
 
 	; AROMA_LADY
 	db "Peony@"
-	db TRAINERTYPE_NORMAL
-
+	db TRAINERTYPE_ITEM | TRAINERTYPE_PERSONALITY
 	; TODO: party
-	dbp 50, BULBASAUR
+	dbp LEVEL_FROM_BADGES + 30, BELLOSSOM, FEMALE
```
itsdarsh, line 8813: I mean this more to be illustrative than to imply this is the only issue: I don't think it's a good idea to have a random trainer with a Level 98 Bellossom.

itsdarsh:
If you could separate your move fixes from the trainer and wild pokemon changes, I can merge those right away. Thanks for fixing those!

data/trainers/parties.asm
katedering: That /one/ specific trainer may be a bit high yes, but, again, overall the level curve is about the same as vanilla if you go there right upon unlocking Kanto.

katedering:
what would be the easiest way to do that, close this and put it up as 2 different PRs?

katedering:
Closing for now, will PR the moves.asm commit as a separate one.

PR #925, opened 2024-01-27
Fix #902 pokedex keeps last visited mon

### LINKED ISSUE #902: (9bit) Newdex doesn't remember last viewed 'mon.
Vanilla dex did, was this an intentional change? It makes hunting for roamers that much more annoying when you have to manually page all the way through 250+ entries to get to them.

nickjwilde:
This doesn't keep position after using search which I believe would be expected behavior

### DIFF
--- a/engine/pokedex/pokedex.asm
+++ b/engine/pokedex/pokedex.asm
@@ -1,13 +1,7 @@
 Pokedex:
-	xor a
-	ld [wPokedex_CursorPos], a
-	ld [wPokedex_Offset], a
 	call StackDexGraphics
 
 	call Pokedex_Main
-
-	xor a
-	ld [wPokedex_CursorPos], a
 
 	jmp Pokedex_MainLoop
 


FredrIQ:
This fix is invalid, because that WRAM region can be used by other things. That's why I zero it on initialization.

nickjwilde:
If you get a sec can you explain a little more? I'm new to assembly for gbc. 

It worked during play testing but I had a feeling there was some other side effect by not zeroing it out each time.

The only other way I can think of to fix it is saving the previous position in a different wram byte and possibly implementing a setting for toggling saving the positions for last visited?

I would appreciate your insight and if there's a better way to fix it I'd love to help out. I was trying to look for issues that were small but possibly easy to fix.

FredrIQ:
Sorry, didn't see this message until now. I will elaborate a bit.

The WRAM location where wPokedex_CursorPos (and wPokedex_Offset) is located is in a section union. What this means is that other features of the game will reuse the RAM that the pokédex uses for other purposes as noted from the same SECTION UNION name ("Misc 404"). For this reason, I need to clear that data during pokédex initialization.

This doesn't mean that this feature can't be implemented, just that the particular way this pull request goes about it isn't really proper. What should be done instead (and what I intend to do) is to use some other WRAM to store the cursor data instead, some location that is unused.

nickjwilde:
@FredrIQ - Thanks for the explanation. That clears it up. If you don't mind I'd like to take a stab at fixing it properly then.

I'll update and reopen if I get to it before you do.

PR #926, opened 2024-01-28
Optimize finding an empty pal slot

vulcandth:
Just a small optimization to finding an empty pal slot.


Rangi42 (commented):
Is `[wUsedObjectPals]` definitely not already full `%11111111`? If it were, this would infinitely loop.

vulcandth:
> Is `[wUsedObjectPals]` definitely not already full `%11111111`? If it were, this would infinitely loop.

Oh poop. Your right. Good catch.

engine/gfx/dynamic_pals.asm
```diff
@@ -147,21 +147,19 @@ MarkUsedPal:
...
-	ld d, a
-.search_again
-	ld a, d
-	and [hl]
-	jr z, .found_empty
-	ld a, d
-	rla
-	ld d, a
+	ld a, [wUsedObjectPals]
+	cp %11111111
```
Rangi42, line 151: You can do `inc a :: jr z, .none_available :: dec a` ([wiki](https://github.com/pret/pokecrystal/wiki/Optimizing-assembly-code#compare-a-to-255)). Will think about the loop later; should be possible to use the carry flag instead of `bit 0` checks.

Rangi42 (commented):
```asm
	push bc
	; Pal is not already loaded, find a empty pal slot
	ld a, [wUsedObjectPals]
	ld b, -1
	scf ; clear carry flag, so even if a == %11111111,
	ccf ; the .bit_check_loop will still terminate with b == 8
.bit_check_loop
	inc b
	rrca
	jr c, .bit_check_loop
	ld a, b
	pop bc
```

This would end up using a nonexistent 8th slot if there was no room, but that is afaict what it did already. If it really should use the 7th slot in that case, use this:

```asm
	push bc
	; Pal is not already loaded, find a empty pal slot
	ld a, [wUsedObjectPals]
	inc a
	jr nc, .some_available
	ld b, 7
	jr .unset_bit_found
.some_available
	dec a
	ld b, -1
.bit_check_loop
	inc b
	rrca
	jr c, .bit_check_loop
.unset_bit_found
	ld a, b
	pop bc
```

(Untested.)

vulcandth:
```diff
	; Pal is not already loaded, find a empty pal slot
	ld a, [wUsedObjectPals]
	inc a
-	jr nc, .some_available
+	jr nz, .some_available
	ld b, 7
	jr .unset_bit_found
.some_available
	dec a
	ld b, -1
.bit_check_loop
	inc b
	rrca
	jr c, .bit_check_loop
.unset_bit_found
	ld a, b
	pop bc
```

Small tweak to your second code suggestion to prevent infinite loop. 

I should have just used the 7th slot (in my original code, not the non-existent 8th), as using the 8th would cause `wLoadedObjPal0` to overflow.. marking the wrong slot in `wUsedObjectPals`. I tested too many unique pals on the screen, and it just causes the second to last npc to be on the screen to have the wrong pal... Which is better than infinite loops.

PR #927, opened 2024-01-30
Fix #899 auto scroll set to start closes move description.

### LINKED ISSUE #899: Instant text speed + START autoscroll bug
https://github.com/Rangi42/polishedcrystal/assets/26468141/f833d6b5-1a3b-4f8e-9fcd-0e285531ae51

Viewing a move's effect in battle with START is glitched when setting text speed to instant and autoscroll to START, in that it instantly closes out. Video proof attached.

nickjwilde:


FredrIQ:
I don't really like this fix. Disabling autoscroll out of the blue feels inconsistent and confusing UI-wise. What do others think?

PR #954, opened 2024-04-27
Allow holding B on cycling road to brake the bike

etdv-thevoid:
Acts similar to the Gen I / Gen III remake behavior on Cycling Road. Holding B will brake the bike while going downhill on Cycling Road.


Rangi42:
Thank you!

PR #959, opened 2024-05-19
Objects may be retained through map connections

vulcandth:
If an object's xy coordinates relative to the player's matches a new mapobject's xy coordinates relative to the player, and their sprite matches, then than object will be retained/reassociated with the new mapobject.

This resolves the issue of object reshuffling when crossing map connections; particularly the cut tree flicker on route32.


engine/gfx/color.asm
```diff
@@ -935,7 +935,14 @@ LoadMapPals:
...
 
 .got_pals
+	ld hl, wPalFlags
+	bit MAP_CONNECTION_PAL_F, [hl]
+	jr nz, .skip_clearing_obj_pals
 	farcall ClearSavedObjPals
+.skip_clearing_obj_pals
+
+	ld hl, wPalFlags
+	res MAP_CONNECTION_PAL_F, [hl]
```
Rangi42, line 945: Would this work instead?
```suggestion
	ld hl, wPalFlags
	res MAP_CONNECTION_PAL_F, [hl]
	bit MAP_CONNECTION_PAL_F, [hl]
	jr nz, .skip_clearing_obj_pals
	farcall ClearSavedObjPals
.skip_clearing_obj_pals
```

Rangi42, line 945: Would this work instead? Note that `res` does not disturb the flags set by `bit`.
```suggestion
	ld hl, wPalFlags
	bit MAP_CONNECTION_PAL_F, [hl]
	res MAP_CONNECTION_PAL_F, [hl]
	jr nz, .skip_clearing_obj_pals
	farcall ClearSavedObjPals
.skip_clearing_obj_pals
```

engine/overworld/map_objects.asm
```diff
@@ -6,11 +6,16 @@ DeleteMapObject::
...
+	push bc
 	ld h, b
 	ld l, c
 	ld bc, OBJECT_LENGTH
 	xor a
 	rst ByteFill
+	pop bc
+	ld hl, OBJECT_MAP_OBJECT_INDEX
+	add hl, bc
+	ld [hl], -1
```
Rangi42, line 18: Would this work instead?
```suggestion
	add hl, bc
	push hl
	ld a, [hl]
	push af
	ld h, b
	ld l, c
	ld bc, OBJECT_LENGTH
	xor a
	rst ByteFill
	pop hl
	ld [hl], -1
```

engine/overworld/map_setup.asm
```diff
@@ -69,12 +69,173 @@ SuspendMapAnims:
...
+	ld hl, OBJECT_LAST_MAP_X
+	add hl, bc
+	ld a, [hli]
+	ld e, [hl]
+	ld d, a
+	call .CheckForMatchingMapObject
+	push af
+	call nc, .DeleteObjectStruct
+	pop af
+	jr nc, .continue
```
Rangi42, line 98: How about this...
```suggestion
	jr nc, .DeleteObjectStruct
```

engine/overworld/map_setup.asm
```diff
@@ -69,12 +69,173 @@ SuspendMapAnims:
...
+	xor a
+.delete_loop
+	ld [hli], a
+	dec d
+	jr nz, .delete_loop
+	; mark the object as free.
+	ld hl, OBJECT_MAP_OBJECT_INDEX
+	add hl, bc
+	ld [hl], -1
+	ret
```
Rangi42, line 147: ...plus this?
```suggestion
	jr .continue
```

engine/overworld/player_object.asm
```diff
@@ -136,13 +136,13 @@ CopyObjectStruct::
 	call CheckObjectMask
 	and a
 	ret nz ; masked
-	ld hl, wObjectStructs + OBJECT_LENGTH * 1
+	ld hl, wObjectStructs + (OBJECT_LENGTH * 1) + OBJECT_MAP_OBJECT_INDEX
```
Rangi42, line 139: Why was that `* 1` even there...
```suggestion
	ld hl, wObjectStructs + OBJECT_LENGTH + OBJECT_MAP_OBJECT_INDEX
```

engine/overworld/map_setup.asm
vulcandth: We want to continue executing the loop.. not jump to delete and return out of the entire routine.

engine/overworld/map_objects.asm
```diff
@@ -6,11 +6,14 @@ DeleteMapObject::
 	add hl, bc
 	ld a, [hl]
 	push af
+	push hl
```
Rangi42, line 9: I think instead of `push hl / ... / pop hl / ld [hl], -1`, you can do `ld [hl], -1 / inc hl` and change `ld bc, OBJECT_LENGTH` to `ld bc, OBJECT_LENGTH - 1`.

vulcandth: We couldn't do exactly that.. but I was able to optimize it some I think.

vulcandth:
Either merge this _before_ weather.. or _after_ weather.. Do not merge at the same time... I'll need to verify that the changes they both do to the object system do not conflict... so we should delay one of these PR's merge so I can verify there are not conflicts.

vulcandth:
Resolved merge conflicts; drafting this until I feel comfortable there are no in game conflicts with `weather`.

Issue #905, opened 2023-11-30
cosmetic 9bit issues


itsdarsh:
Will be expounded upon as I continue my playthrough

- [x] if a pokemon with a stone evolution learns a move post-evolution, the screen returns to the party menu while learning the move (it should stay on a white screen, otherwise you see the whole party with Not Able listed next to it, which looks confusing imo)
- [x] if there's a weather condition and you're out of moves, the "has no moves left" message appears under the weather icon
- [x] conversely, pulling up a move description doesn't redraw the weather icon until you go back to the main battle menu
- [x] the withdrawing into ball anim plays twice for Pursuit used on a mon switching out if battle animations are off'
- [x] moves that never miss (or status moves that otherwise ignore accuracy checks) are listed with 100% accuracy, not ---% accuracy
- [x] in Goldenrod Harbor, the "kaching" sound effect when buying something doesn't match when money is actually reduced
- [x] After the Bug Catching Contest is over due to the player opting to end it early, the BG fades in from black, but the NPC sprites do not, meaning a second or two where the NPCs are all standing on a black void before the BG fades in
- [x] If enemy mon has a max length name (e.g. Jigglypuff) and uses a palette changing animation (for example, eating a berry), the palette of the last letter will also change with the animation
- [x] Leaving the Route 36 grotto puts your spriteover the grass awkwardly
![image](https://github.com/Rangi42/polishedcrystal/assets/55897426/86acac23-9cc6-43d8-8254-128f9718ba54)

PR #962, opened 2024-05-30
Fix grotto exit on route36 player standing on grass

vulcandth:
Player will no longer be standing "above" the grass when exiting the grotto into grass. If there are other grottos that may exit onto grass, you simply need to call `UpdatePlayerTallGrassFlags` from the maps OBJECT callback.

Reference #905

### DIFF
--- a/data/maps/setup_script_pointers.asm
+++ b/data/maps/setup_script_pointers.asm
@@ -59,3 +59,4 @@ MapSetupCommands:
 	add_mapsetup SetCurrentWeather ; 34
 	add_mapsetup MapConnOWFadePalettesInit ; 35
 	add_mapsetup LoadMapObjects_Connection ; 36
+	add_mapsetup GrottoUpdatePlayerTallGrassFlags ; 37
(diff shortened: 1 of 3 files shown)


maps/Route36.asm
```diff
@@ -48,6 +48,7 @@ Route36_MapScriptHeader:
 	const ROUTE36_COOLTRAINERF
 
 Route36ArthurCallback:
+	callasm UpdatePlayerTallGrassFlags
```
Rangi42, line 51: Can/should this just be done unconditionally when you exit any grotto?

vulcandth: Depends on what you mean... You can create an object callback with that line on every map that has a grotto...

Or... I could include it as a mapscript... and it would do it upon *every* map load.. Which would work, and *shouldn't* cause any problem... but I was trying to avoid make a wide spread change for what appeared to be a single instance of a problem. Would you prefer the mapscript option?

Rangi42: A mapscript would be fine. I think exiting a grotto could set a flag to true; then the new mapscript would check this flag, just return if it's false, but if it's true then do the fix and set the flag to false.

PR #969, opened 2024-06-30
Boulders block trainer sight

### LINKED ISSUE #381: Long-distance trainers are not blocked by Strength boulders
image

vulcandth:
I wouldn't call this _efficient..._ but it technically works. This will prevent trainer sight when a boulder is in the way. If we removed the boulder sprite check it will prevent trainer sight if _anything_ is in the way.  
  
My brain doesn't want to be efficient right now, so If you see ways to optimize please suggest and i'll apply. Furthermore, we may want to farcall some of this so it's not in the home bank? I'll leave that up to you.  
  
Resolves #381

### DIFF
--- a/home/trainers.asm
+++ b/home/trainers.asm
@@ -229,12 +229,219 @@ FacingPlayerDistance::
 	call GetSpriteDirection
 	cp e
 	jr nz, .NotFacing
+
+	push de
+	ld hl, OBJECT_MAP_X ; x
+	add hl, bc
+	ld d, [hl]
+
+	ld hl, OBJECT_MAP_Y ; y
+	add hl, bc
+	ld e, [hl]
+
+	cp OW_LEFT
+	jr z, .check_boulder_left
+	cp OW_RIGHT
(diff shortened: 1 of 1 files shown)


home/trainers.asm
```diff
@@ -229,13 +229,220 @@ FacingPlayerDistance::
...
+	add hl, bc
+	ld a, [hl]
+	cp e
+	jr nz, .next_left_check
+	; check for boulder sprite
+	ld hl, OBJECT_SPRITE
+	add hl, bc
+	ld a, [hl]
+	cp SPRITE_BOULDER_ROCK_FOSSIL
+	jr nz, .next_left_check
```
Rangi42, line 284: I think *if* we do this at all, it should be a general-purpose "no trainer sight if there's any NPC in the way" check. (And even then, I'd prefer to redesign maps so the circumstance can't occur, if possible.)

Rangi42:
Tested and this works with Cinnabar Volcano's Strength boulders. No visible performance lag in the overworld. :+1:

PR #973, opened 2024-07-13
Shift Nuzlocke Icon Left, Don't skip caught icon

vulcandth:
- Shift Nuzlocke icon left a square.
- Don't skip the caught icon is the Nuzlocke icon is showing

We'd probably have to re-arrange other things if we don't want to shift this far left. See how it looks now below.

image

It being shifted left doesn't look bad when the caught icon is *also* showing.. but when its just the nuzlocke icon, it can look a little off.. I could adjust the code to only shift the nuzlocke icon if the caught icon *isn't* showing. Let me know what you think.

### DIFF
--- a/engine/gfx/cgb_layouts.asm
+++ b/engine/gfx/cgb_layouts.asm
@@ -282,6 +282,7 @@ _CGB_FinishBattleScreenLayout:
 	call FillBoxWithByte
 
 	ld a, PAL_BATTLE_BG_EXP_GENDER
+	ldcoord_a 0, 1, wAttrmap
 	ldcoord_a 1, 1, wAttrmap
 	ldcoord_a 8, 1, wAttrmap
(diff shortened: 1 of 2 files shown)


engine/gfx/cgb_layouts.asm
```diff
@@ -271,6 +271,10 @@ _CGB_FinishBattleScreenLayout:
 	ld a, PAL_BATTLE_BG_ENEMY_HP
 	call FillBoxWithByte
 
+	; set nuzlocke icon palette
+	hlcoord 0, 1, wAttrmap
+	ld [hl], PAL_BATTLE_BG_EXP_GENDER
+
```
Rangi42, line 277: This should be one `ldcoord_a 0, 1, wAttrmap` line below, along with the other three.

Rangi42 (approved):

PR #974, opened 2024-07-13
Update Readme screenshots

vulcandth:
This PR updates all of the current `master` screenshots to their equivalents in `9bit`. We can add some screenshots of newer features later if we want... this just updates what is already there...

We still have the following screenshots that need updating....
- [x] custom_map.png (I don't know where this is)
- [x] sitrus-berry.png (I don't know where this is)
- [ ] togetic-fairy-wing.png (Fairy wind doesn't exist anymore)
- [ ] new-balls.png (I don't know where this mart is)


FrenchOrange:
"custom_map" is Yellow Forest I believe?
"sitrus_berry" is Route 45.
"new-balls" doesn't seem to be an actual mart lineup?

vulcandth:
I suggest we just don't update those last two.. and use those slots to add some new screenshots later.

screenshots/eviolite.png
Rangi42, line 1: "Aaaaaaa" could be better :P

Rangi42:
Some of these really highlight how different standards were in 2017 or whenever. (Two Town Map pics? Two "X wants to battle" pics? A pic just to illustrate that yes, we have wild Remoraid now?) But as a simple update to the old ones, this is great; now they can be replaced without having to worry whether the reminder are subtly outdated. Thank you!

PR #977, opened 2024-08-08
Load Weather Icon sprite if present on MoveDescription close

walshyb:
## Issue
Per #905: pulling up a move description doesn't redraw the weather icon until you go back to the main battle menu

## Before Example

https://github.com/user-attachments/assets/aef3c615-5e3e-47a5-8271-2c38eb31c59e

## Fixed Example

https://github.com/user-attachments/assets/65d1c59c-c3e9-4981-a609-8f5bdbc0c02a

I'm not sure how testing / validation works for this. Please let me know if I'm missing anything, thank you!


vulcandth:
Thanks! This looks good; probably leave this open until I can do some quick testing myself and I can possibly add the fix for "if there's a weather condition and you're out of moves, the "has no moves left" message appears under the weather icon" into this PR as well. I should have time to do that either tonight, or this weekend.

walshyb:
Totally, thank you so much! If I have some time I'll see if I can figure out that other issue. But it's also fine if you get to it first!

walshyb:
@vulcandth I think I got it working! Seems so at least when I tested. Here's the vid: 

https://github.com/user-attachments/assets/608d6dd1-cb2f-42ae-8e40-68cda09e2673

Notes / questions:
- Had to change the `jr .start_over` to `jp .start_over` because having the `LoadWeatherIconSprite` call in `.pressed_start` as well as the `ClearSprites` call in `.struggle` gave me: `jr target must be between -128 and 127 bytes away, not -129; use jp instead`
- Haven't when an enemy pokemon uses struggle; proabably worth testing to ensure no regression
- Do you think that's a good place to call `ClearSprites`?

engine/battle/core.asm
```diff
@@ -5201,7 +5202,8 @@ MoveSelectionScreen:
 	pop hl
 	call BattleMoveDescTextbox
 	call WaitPressAorB_BlinkCursor
-	jr .start_over
+	call LoadWeatherIconSprite
+	jp .start_over
```
vulcandth, line 5206:
```suggestion
	jmp .start_over
```
Polished Crystal uses `jmp` instead of `jp`. `jmp` is a macro that evaluates to a normal `jp` except that it also comes with an assertion that will warn during build time if it ever can be optimized back to being a `jr`.

engine/battle/core.asm
```diff
@@ -5173,6 +5173,7 @@ MoveSelectionScreen:
 	jmp MoveSelectionScreen
 
 .struggle
+	call ClearSprites
```
vulcandth, line 5176: This is typically fine to use; but keep in mind it clears **_ALL_** sprites. I think when you are calling this, there are no sprites on screen other than the weather image. Unless someone thinks of an edge case; should be fine.

vulcandth (changes requested):

PR #987, opened 2025-01-02
Add Event Flag Checksums

vulcandth:
Example of an active event checksum.

- Calculates the summation checksum (modulo 256) of the entire event space and stores it if not done so already.
- Recalculates the checksum when we modify any event flag.
- VBlank routine calculates one byte at a time (~5 secs total to do all event bytes) and then crashes on mismatch.


vulcandth:
Will not implement now; but this is a PoC for those interested.. the branch on my fork will remain for sometime in the future.

PR #1001, opened 2025-01-29
Store abnormal contact moves instead of all contact moves

emilyploszaj:
Minor size and runtime optimization. This should also help not mistakenly forget to mark a typical physical move as contact, as the property seems non-notable.

Moves typically make contact if they're physical and don't if they're special. Polished only has 21 moves with abnormal contact interactions, as opposed to the 92 moves which make contact. This adjustment costs 16 bytes of code to save 70 bytes of table space, with an overall net 54 bytes saved. Future efforts towards 10bit moves or similar will retroactively double this saved space.

I tested this code change on a rocky helmet mon using draining kiss, bonemerang, tackle, psychic, and struggle. All moves behaved as expected (draining kiss, tackle, and struggle made contact, bonemerang and psychic did not make contact).


FredrIQ (changes requested):
We never write out `a` if implicit

Issue #1006, opened 2025-02-01
Bloodmoon Ursaluna can't learn Earth Power via MT


erejordan:
Not sure if this is a bug or a balancing choice but Bloodmoon Ursaluna cannot learn Earth Power via MT.

NickF56:
Hijacking this issue to also flag Exeggutor not having calm mind access so it’s not lost

futuresushi:
RE: Canonical "should learns", I made a big dumb spreadsheet about it here:
https://docs.google.com/spreadsheets/d/1lbZTAANClak8fay6ZnVDmHgnY9ifpzvp5NVbzXfafQA/edit?usp=sharing

My methodology is explained on the About page, open to comments. Hope this is helpful!

itsdarsh:
I'd like to go through that spreadsheet when I get the chance so it may become its own issue. This specific issue is closed.

PR #1020, opened 2025-02-04
Fixing The Sins Of Adding A Cool Mon And Making It Bad

TAG1Kilo:
Bloodmoon Ursaluna does not learn Earth Power in the current version of Polished Crystal in any way, shape, or form. According to Bulbapedia it learns it by level up and by TM. It is literally its best Ground type STAB move, as it's a special attacker. ...But it doesn't learn it. At all. WHY!? (Incidentally, regular Ursaluna can learn Earth Power by TM too so I added it to the MT list as well for both to reflect that. The level I chose for BM Ursa to learn the move naturally is very close to when it learns it in mainline games.)


earthoul:
You'll also want to add Earthquake to Ursaluna's level-up moves as a counterpart of Ursaluna BM's Earth Power (#1006)

PR #1043, opened 2025-02-14
Use updated sprite designs by Avoman

Rangi42:
Supersedes #904

Adds most of the sprites from https://github.com/avoman95/polishedcrystal/tree/new-sprites, except for Togepi's back sprite (it's small like Cleffa and Igglybuff, so its whole body + feet are visible).

Also I edited the Crys back sprite to be a few pixels taller so their head matches the others.

Requesting a review from SourApple in case any designs conflict with other new/upcoming ones from https://github.com/SoupPotato/pokecrystal_SpriteBase/.


SoupPotato:
Hmm, Im not a fan of most the back sprites, especially Blastoise. 
image

The perspective of the cannon's on the original sprite were very off. I have corrected them in the new backsprite

[image](https://github.com/Rangi42/polishedcrystal/blob/SourSprites-9bit/gfx/pokemon/blastoise/back.png?raw=true)

Same thing for most of them. Such as metapod which lacks significant shading:

image

My one:

https://github.com/Rangi42/polishedcrystal/blob/SourSprites-9bit/gfx/pokemon/metapod/back.png?raw=true


Im sorry, but I just dont feel like these ones do very much justice.

SoupPotato:
Actually, this is the original Metapod backsprite from Crystal:

image

The one polished has used to until now is custom.

As for the other sprites, I mean its your hack so its your call :P

avoman95:
I've updated some of the back sprites mentioned here on my branch over the last few days for a more gen 2 style of shading, plus an extra jumpluff sprite.

itsdarsh:
I have an incoming new Suicune back from @ShadowOne333 (I think it does the same thing as the one in this PR) so just be aware.

avoman95:
I Believe Soup has also made a new improved chinchou back now, which I think is better.

avoman95:
Made a few adjustments to the crys back sprite. Raised the hairline and made the spikes cleaner, smoothed out the shoulders and other jagged parts. Overall, a lot happier with it now. @Rangi42 would you like to use this one instead?
crys_back
crys_back_colour

PR #1045, opened 2025-02-15
Rename "items" comment to "held items" in base stats files

walkawayy:
Only Goldeen, Seaking, and Smoochum were commented differently for some reason. This is just a cosmetic difference I noticed while parsing the files for the wiki.


itsdarsh:
That's my bad, I probably forgot to change it after reintroducing them in 9bit :V

PR #1046, opened 2025-02-15
Change UrsalunaBloodmoonEvoAttacks to UrsalunaBloodmoonEvosAttacks

walkawayy:
Another difference I noticed while parsing for the wiki. Bloodmoon Ursaluna's moves weren't getting parsed. For some reason it was the only Pokemon that used `EvoAttacks` instead of the plural `EvosAttacks`. I believe this should only be cosmetic but let me know if it's not.


itsdarsh:
Thanks!

Issue #1073, opened 2025-02-28
radios begin with garbage text


itsdarsh:
![Image](https://github.com/user-attachments/assets/66e095f3-6dc3-440e-b76b-9c2341abd4e7)

Found by Dia on the discord. She found that commit 5b8306b09d06b8d177901a06da15a908d3b9218b is the first commit that affects relevant data, so it could be an issue with the update to text compression. Haven't looked into it yet.

PR #1079, opened 2025-03-01
Decompress radio text before normal radio logic

vulcandth:
1. Copy radio line into `wRadioCompressedText`
2. Check if text is compressed.
3. If compressed, decompress `wRadioCompressedText + 1` into `wRadioText + 1`, normal copy otherwise. (We assumed the first byte is `<START>` and the second byte may be `<CTXT>` if compressed)


vulcandth:
Related to #1073

PR #1122, opened 2025-03-20
Optimize `cp 1` to `dec a` when possible

Rangi42:
There's one `cp 1` in engine/battle/effect_commands.asm that might be able to be `dec a`, I didn't bother checking all the call sites.

### DIFF
--- a/engine/battle/effect_commands.asm
+++ b/engine/battle/effect_commands.asm
@@ -5005,11 +5005,13 @@ CanStatusTarget:
 .pop_and_end
 	pop af
 .end
+	; return nc|nz -- we can't, failure msg in HL
 	or 1
 	ret
 .cant_ability
-	xor a
-	cp 1 ; no-optimize a == 1 (TODO: can this be dec a?)
+	; return c|nz -- we can't, due to ability
+	scf
+	sbc a
 	ret
 
 BattleCommand_poisontarget:
--- a/engine/gfx/sprite_anims.asm
+++ b/engine/gfx/sprite_anims.asm
@@ -81,18 +81,14 @@ AnimSeq_PartyMonSwitch:
 	add hl, bc
 	ld a, d
 	and $10 ; bit 4
+	; This works because a>1, meaning zero is always written to [hl]
 	jr z, .load_zero
 	ld a, e
-	and a ; 0?
-	jr z, .load_minus_two
-	dec a ; 1?
-	jr z, .load_minus_one
 .load_zero
-	ld a, 2
-.load_minus_two
-	dec a
-.load_minus_one
-	dec a
+	sub 2
+	jr c, .got_load
+	xor a
+.got_load
 	ld [hl], a
 	ret
 
--- a/engine/menus/naming_screen.asm
+++ b/engine/menus/naming_screen.asm
@@ -573,7 +573,6 @@ NamingScreen_GetDPad:
 	ld a, $3
 .wrap_around_command_left
 	dec a
-	dec a
 	ld e, a
 	add a
 	add e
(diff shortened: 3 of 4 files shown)


engine/battle/effect_commands.asm
```diff
@@ -5009,7 +5009,7 @@ CanStatusTarget:
 	ret
 .cant_ability
 	xor a
-	cp 1
+	cp 1 ; no-optimize a == 1 (TODO: can this be dec a?)
```
FredrIQ, line 5012: The point of this code is to return c|nz, see the return states at the top of the function. So no, it can't be replaced with `dec a` since that doesn't set carry. This, however, would work:
```diff
 .cant_ability
-	xor a
-	cp 1
+	scf
+	sbc a
	ret
```

engine/events/engine_flags.asm
```diff
@@ -25,20 +25,19 @@ EngineFlagAction::
...
 ; What are we doing with this flag?
 	ld a, b
-	cp 1
-	jr c, .reset ; b = 0
-	jr z, .set   ; b = 1
+	and a ; 0?
+	jr z, .reset
+	dec a ; 1?
+	ld a, [de]
+	jr z, .set
```
FredrIQ, line 32: `ld a, [de]` optimization aside, isn't the former more readable? And you don't really save space by doing `and a [...] dec a`.

engine/gfx/sprite_anims.asm
```diff
@@ -83,21 +83,17 @@ AnimSeq_PartyMonSwitch:
...
-	ld [hl], -1
-	ret
-
+	ld a, 2
 .load_minus_two
-	ld [hl], -2
+	dec a
+.load_minus_one
+	dec a
+	ld [hl], a
```
FredrIQ, line 96: This can be optimized further as such:
```diff
@@ -81,25 +81,18 @@ AnimSeq_PartyMonSwitch:
	add hl, bc
	ld a, d
	and $10 ; bit 4
+
+	; This works because a>1, meaning zero is always written to [hl].
	jr z, .load_zero
	ld a, e
-	and a
-	jr z, .load_minus_two
-	cp 1
-	jr z, .load_minus_one
 .load_zero
+	sub 2
+	jr c, .got_load
	xor a
+.got_load
	ld [hl], a
	ret

-.load_minus_one
-	ld [hl], -1
-	ret
-
-.load_minus_two
-	ld [hl], -2
-	ret
-
 AnimSeq_PartyMonSelected:
        ld a, [wMenuCursorY]
```

engine/items/mart.asm
```diff
@@ -585,9 +585,9 @@ MartAskPurchaseQuantity:
 	inc hl
 	inc hl
 	ld a, [hl]
-	and a
+	and a ; 0?
 	jmp z, StandardMartAskPurchaseQuantity
-	cp 1
+	dec a ; 1?
```
FredrIQ, line 590: `BargainShopAskPurchaseQuantity` immediately sets a to 1 (and isn't called elsewhere), so you could keep the `cp 1` and remove the `ld a, 1` from it.

engine/menus/naming_screen.asm
```diff
@@ -568,9 +568,9 @@ NamingScreen_GetDPad:
 	ret
 
 .caps_del_done_left
-	cp $1
+	dec a
 	jr nz, .wrap_around_command_left
-	ld a, $4
+	ld a, $3
 .wrap_around_command_left
 	dec a
```
FredrIQ, line 575: Did you mean to remove a `dec a` here?

FredrIQ (commented):
Some things could be optimized further, and I found one location that is almost certainly a mistake. Looks fine otherwise.

PR #1124, opened 2025-03-21
[10bit] Expand move anims and add a handful of moves

emilyploszaj:
First part, I expanded the anim code to allow multiple banks to be used. Could be optimized with an indirection table, we're *really* short on rom0 though so for now this'll work. I moved substitute bypass moves to make space because it's only used in a single location.

Second part, moves. All tested to like, probably mostly work. Move choices were really just what I was feeling, I added some collapsed "duplicate" moves back (though haven't made changes like HEALING_LIGHT -> SYNTHESIS) and some more fun moves. The three elemental fangs are odd moves. They each have a 10% chance to flinch and an *independent* 10% chance to status. Rolling effect chance twice on 10% would probably work but would have some unintended consequences. However, a simple breakdown of these odds has us realize that there's a 1% chance to both flinch and status, a 9% to flinch and not status, and a 9% to status but not flinch. This sums to 19%, so these moves have been given that chance and the effect rolls which of these 19 "cases" happened.


FredrIQ:
Note that my code review is primarily focused on the battle engine aspect, I have not reviewed the animation changes.

FredrIQ (commented):
Most of it looks fine to me, but there are some minor suggestions on tweaks.

data/moves/animation_pointers.asm
```diff
@@ -1,297 +1,308 @@
 ; entries correspond to constants/move_constants.asm
 :
-	table_width 2, :-
+	table_width 3, :-
```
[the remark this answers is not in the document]
vulcandth: This is from code I ported from PC16 and is on me to fix. I'll fix it after this PR.

emilyploszaj:
Comments addressed

vulcandth:
Slightly unrelated to this; but I wonder if we should do something like this:

```diff
	const anim_call_command ; fe
+	const anim_farcall_command ; ff
MACRO anim_call
+	if BANK(\1) == BANK(@)
		db anim_call_command
		dw \1 ; address
+	else
+		db anim_farcall_command
+		dba \1 ; bank address
+	endc
ENDM
```

It would require shifting around the command id's... since there is already a `ff` command ID... But this would keep move animators (sour) from having to worry about if the referenced animation script is far or not... Currently if you do an `anim_call` you have to make sure its in the same file (bank)...

data/moves/animation_pointers.asm
```diff
@@ -1,297 +1,308 @@
...
+	dba BattleAnim_SendOutMon
+	dba BattleAnim_ReturnMon
+	dba BattleAnim_Transform
+	dba BattleAnim_Miss
+	dba BattleAnim_EnemyDamage
+	dba BattleAnim_PlayerDamage
+	dba BattleAnim_HitConfusion
+	dba BattleAnim_HeldItemTrigger
 	assert_table_length NUM_BATTLE_ANIMS
 BattleAnimations::
```
vulcandth, line 33: Maybe we should just make this an indirection table to save bytes on the table. It would look something like this:


These are **Examples** from CSE, obviously would adapt accordingly.
```x86asm
BattleAnimations::
; entries correspond to constants/move_constants.asm
	indirect_table 2, 0
	indirect_entries GEN1_MOVES, BattleAnimationsGen1
	indirect_entries GEN2_MOVES, BattleAnimationsGen2
	indirect_entries $ffff - NUM_BATTLE_ANIMS
	indirect_entries $ffff, BattleAnimationsNegatives
	indirect_table_end


SECTION "Battle Animations Gen 1", ROMX

INCLUDE "data/moves/animations_gen1.asm"


SECTION "Battle Animations Gen 2", ROMX

INCLUDE "data/moves/animations_gen2.asm"

SECTION "Battle Animations Gen Negatives", ROMX

INCLUDE "data/moves/animations_negatives.asm"

ENDSECTION
```


```x86asm
GetBattleAnimPointer::
 	ld hl, BattleAnimations
 	ld a, BANK(BattleAnimations)
 	call LoadDoubleIndirectPointer
 	jr nz, .found
 	ld hl, BattleAnim_Dummy
 .found
 	ld [wBattleAnimBank], a
 	ld a, l
 	ld [wBattleAnimAddress], a
 	ld a, h
 	ld [wBattleAnimAddress + 1], a
 	ret
 	
 	
 	GetBattleAnimByte::
	push hl
	push de

	ld hl, wBattleAnimAddress
	ld a, [hli]
	ld d, [hl]
	ld e, a

	ld a, [wBattleAnimBank]
	rst Bankswitch

	ld a, [de]
	ld [wBattleAnimByte], a
	inc de

	ld a, BANK(BattleAnimCommands)
	rst Bankswitch

	ld a, d
	ld [hld], a
	ld [hl], e

	pop de
	pop hl

	ld a, [wBattleAnimByte]
	ret
```

```
BattleAnimationsGen1::
	dw BattleAnim_Dummy
	dw BattleAnim_Pound
	dw BattleAnim_KarateChop
	...
.IndirectEnd::
```

emilyploszaj: I mention this in the desc, we can definitely do an indirection table but we're so short on ROM 0 space to handle it atm that I'd prefer we leave it as a future optimization if possible

emilyploszaj:
Probably a good idea to add the indirection table, cross bank anim calls, and other anim expansions to be in a separate PR and leave this one for just the new moves

vulcandth:
> Probably a good idea to add the indirection table, cross bank anim calls, and other anim expansions to be in a separate PR and leave this one for just the new moves

Sounds good!

vulcandth (approved):
Looks good to me. If you are satisfied with the feedback you received feel free to merge yourself if you are not waiting for any additional feedback from FIQ or Sylvie.

PR #1216, opened 2025-06-21
Document auto-generated Huffman tree

vulcandth:
## Summary
- Building the ROM after updating the Huffman codes reported Free space: 42404/2097152 (2.02%) showing a slight increase in free space compared to before

## Testing
- `make -s`

------
https://chatgpt.com/codex/tasks/task_e_6856c2bc75308325af7d8222b9c09830

### DIFF
--- a/data/text/compressed_text.asm
+++ b/data/text/compressed_text.asm
@@ -1,3 +1,5 @@
+; Huffman tree nodes are generated from the codes in constants/charmap.asm.
+; Regenerate this table whenever the Huffman codes change.
 DEF num_parent_nodes = ROOT_NODE_ID
 DEF node_prefix_{02X:num_parent_nodes} EQUS ""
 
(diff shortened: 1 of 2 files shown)


data/text/compressed_text.asm
```diff
@@ -1,3 +1,5 @@
+; Huffman tree nodes are generated from the codes in constants/charmap.asm.
+; Regenerate this table whenever the Huffman codes change.
```
Rangi42: It's auto-generated with macros now ^^
```suggestion
```

Rangi42:
Note that the AI figured out how to run utils/huffman.py and use the codes it outputs. Just to be clear that it's not using some secret special code-designing algorithm, just intelligently running the script we have and updating charmap.asm based on its output. (Saves us a lot of copy+pasting!)

PR #1223, opened 2025-06-30
Require RGBDS 0.9.3 for its DMG palette specs

Rangi42:
[This feature](https://rgbds.gbdev.io/docs/v0.9.3/rgbgfx.1#DMG) should prevent issues like this one:
image


Rangi42:
`rgbgfx -vvvvv` verbose logging shows what the issue was:

> ```
> Image colors: [ #000000ff, #999999ff, #dadadaff, #ffffffff, ]
> Color #ffffffff conflicts with another one, not using grayscale sorting
> FATAL: Image contains too many or non-gray colors, not compatible with a DMG palette specification
> ```

RGB colors are put into four grayscale shade "bins", by scaling their range from 0-255 to 0-3 (i.e. divide by 64). The "light gray" #DADADA is so light that it gets grouped in the same bin as #FFFFFF. (And #999999 is also light enough to count as light instead of dark gray.)

PR #1227, opened 2025-07-10
Faster optimize.py

vulcandth:
Produces the same output as before (including the order of the output; verified by a diff checker) when run against pret/pokecrystal and its ~5k matches, except it completes in 26.24 seconds instead of the original 73.84 seconds.

### DIFF
--- a/utils/optimize.py
+++ b/utils/optimize.py
@@ -5,7 +5,6 @@
 Search all .asm files for N code lines in a row that match some conditions.
 """
 
-from __future__ import annotations
 from argparse import ArgumentParser
 from collections import namedtuple
 from pathlib import Path
@@ -13,7 +12,6 @@ from sys import stderr
 
 # Regular expressions are useful for text processing
 import re
-from typing import List, Tuple
 
 # Paired registers are also useful
 PAIRS = dict(sum([[(x, y), (y, x)] for (x, y) in {'af', 'bc', 'de', 'hl'}], []))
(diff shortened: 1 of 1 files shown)


utils/optimize.py
```diff
@@ -631,51 +668,37 @@ def optimize(filename):
...
-				cur_label = Line(i+1, code, comment, text, code)
-			# Remove indentation from code, if any
-			code = code.lstrip()
-			# Normalize whitespace
-			code = re.sub(r'\s+', ' ', code)
-			# Record the line's properties
-			context = cur_label.code if cur_label else ''
-			cur_line = Line(i+1, code, comment, text, context)
+			if (':' in cur_line.code and (cur_line.code[0].isalpha() or cur_line.code[0] == '_')
+					and not cur_line.text[0].isspace()):
```
Rangi42, line 674: I'm suspicious that this condition is still partly redundant.

utils/optimize.py
```diff
@@ -631,51 +668,37 @@ def optimize(filename):
...
-						print(filename, line.num, line.text, sep=':')
+						print(f'{filename}:{line.num}:{line.text}')
 					printed = True
-					printed_this = True
 					prev_lines = []
 					state = 0
 			elif allow_rewind and not skip:
 				# The condition was not met; go back to a previous condition
-				i -= 1
+				i -= rewind_count
```
Rangi42, line 701: I think this is supposed to be 1, but can't find a codebase that would trigger it right now.

Rangi42:
Confirmed that the output with pokecrystal is 100% identical, and a few times faster than before. Thank you!

Rangi42 (approved):

PR #1229, opened 2025-07-13
[10bit] Fifth wave of move additions

emilyploszaj:
Adds 29 relatively high complexity moves compared to previous waves. Additionally fixes a handful of bugs. I'm not sure all of it works perfectly, but it should be pretty close to it. Would love @FredrIQ insight on battle engine changes.


engine/battle/move_effects/psywave.asm
```diff
@@ -0,0 +1,34 @@
+BattleCommand_psywave:
+	ldh a, [hBattleTurn]
+	and a
+	ld a, [wBattleMonLevel]
+	jr z, .got_level
+	ld a, [wEnemyMonLevel]
+.got_level
+	ld c, a
+	ld a, 101
```
FredrIQ, line 9: I would add a comment here noting that the 0-100 inclusive randomization is intended, because my knee-jerk reaction was "surely this can't be right...". For the record, this is indeed the right range, in spite of it being silly.

edit: Misremembered how modern Psywave works (I blame the move being incredibly forgettable beyond RBY weirdness), the 50-150 randomization makes perfect sense.

data/moves/effects.asm
```diff
@@ -2692,3 +2744,328 @@ Incinerate:
...
+	doturn
+	hastarget
+	checkhit
+	checkpriority
+	memento
+	endmove
+
+FocusPunch:
+	checkobedience
+	focuspunch
```
FredrIQ: Focus Punch cannot be a battle command, because it is checked during pre-move checks (CheckTurn in Polished). The consequences of putting it later like this is, among other things, that Confusion and flinch takes priority when it shouldn't. It is 3.5 in https://wiki.xn--rckteqa2e.com/wiki/%E3%82%BF%E3%83%BC%E3%83%B3#2.%E8%A1%8C%E5%8B%95%E9%81%B8%E6%8A%9E (under "Skill success check" when machine translated in chrome).

engine/battle/move_effects/taunt.asm
```diff
@@ -0,0 +1,32 @@
...
+.order_done
+	ld a, [hl]
+	and $F0
+	jr nz, .failed
+	ld a, [hl]
+	or a, b
+	ld [hl], a
+	ld hl, TauntEffectText
+	farcall AnimateCurrentMove
+	jmp StdBattleTextbox
```
FredrIQ: You need to CheckOpponentMentalHerb here (and make the function capable of healing Taunt).

engine/battle/move_effects/torment.asm
```diff
@@ -0,0 +1,19 @@
...
+	call GetBattleVarAddr
+	bit SUBSTATUS_TORMENTED, a
+	jr nz, .failed
+
+	set SUBSTATUS_TORMENTED, a
+	ld [hl], a
+	
+	ld hl, TormentEffectText
+	farcall AnimateCurrentMove
+	jmp StdBattleTextbox
```
FredrIQ: Torment, like Taunt, is also covered by Mental Herb.

engine/battle/core.asm
```diff
@@ -193,6 +193,14 @@ BattleTurn:
 	inc a
 	ldh [hBattleTurn], a
 	ld [wEnemyGoesFirst], a
+
+	; Display tightening focus in speed order
+	call .check_focus
+	call SwitchTurn
+	call .check_focus
+	call SwitchTurn
```
FredrIQ, line 203: Strictly speaking, Focus Punch's tightening of focus has a priority bracket of +6. This means that switching (+10 in Polished) and Prankster Helping Hand (which we don't have) can beat it to the punch.

I don't think this difference has any practical effects, so if it seems too complicated to handle focus after switching, feel free to ignore it (but you should probably add a comment about it).

engine/battle/move_effects/fury_cutter.asm
```diff
@@ -0,0 +1,32 @@
...
+	dec [hl] ; cap at 255
+.incremented
+	ld b, [hl]
+	ld a, BATTLE_VARS_MOVE_POWER
+	call GetBattleVarAddr
+	ld a, [hl]
+.loop
+	dec b
+	jr z, .done
+	sla a
```
FredrIQ: add a is better

data/moves/effects.asm
```diff
@@ -2692,3 +2744,328 @@ Incinerate:
...
+	endmove
+
+Memento:
+	checkobedience
+	usedmovetext
+	doturn
+	hastarget
+	checkhit
+	checkpriority
+	memento
```
FredrIQ, line 3021: `Use loweroppstat $10 | ATTACK` (and the same with Special Attack) here instead of handling stat lowering through asm.

engine/battle/move_effects/memento.asm
```diff
@@ -0,0 +1,25 @@
...
+
+	farcall CheckSubstituteOpp
+	jr nz, .failed
+
+	;farcall AnimateCurrentMove
+
+	ld b, $10 | ATTACK
+	farcall ForceLowerOppStat
+	ld b, $10 | SP_ATTACK
+	farcall ForceLowerOppStat
```
FredrIQ, line 14: Is there a particular reason as to why you are using asm to do this rather than move commands? `loweroppstat` should work fine as per the effects.asm comment above.

engine/battle/move_effects/psywave.asm
```diff
@@ -0,0 +1,34 @@
...
+	call BattleRandomRange
+	add a, 50
```
FredrIQ: add 50

engine/battle/move_effects/spikes.asm
```diff
@@ -28,15 +28,34 @@ BattleCommand_toxicspikes:
...
 	ld hl, ToxicSpikesText
 	jmp StdBattleTextbox
 
-.failed
-	jmp FailSpikes
+BattleCommand_stickyweb:
+	ldh a, [hBattleTurn]
+	and a
+	ld hl, wEnemyHazards
+	jr z, .got_screens
```
FredrIQ: The label should probably be changed since this has nothing to do with screens.

engine/battle/move_effects/stored_power.asm
```diff
@@ -0,0 +1,60 @@
...
+	ld hl, wPlayerStatLevels
+	ldh a, [hBattleTurn]
+	and a
+	jr z, .ok  
+	ld hl, wEnemyStatLevels
+.ok
+	lb bc, 0, 7
+.loop
+	ld a, [hli]
+	sub a, BASE_STAT_LEVEL
```
FredrIQ: sub BASE_STAT_LEVEL

engine/battle/move_effects/stored_power.asm
```diff
@@ -0,0 +1,60 @@
...
+	or [hl]
+	ret z
+
+	ld hl, wPlayerStatLevels
+	ldh a, [hBattleTurn]
+	and a
+	jr z, .ok  
+	ld hl, wEnemyStatLevels
+.ok
+	lb bc, 0, 7
```
FredrIQ, line 13: You could `lb bc, 1, 7` and then simply not bother returning early if we have no positive stat boosts, since 20 x 1 is still 20. Also removes the need to increment the multiplier since it'd start at x1 already.

engine/battle/move_effects/trump_card.asm
```diff
@@ -0,0 +1,35 @@
...
+	jr z, .got_move_num
+	ld a, [wCurEnemyMoveNum]
+	ld hl, wEnemyMonPP
+.got_move_num
+	add l
+	ld l, a
+	adc h
+	sub l
+	ld h, a
+	; got pp
```
FredrIQ, line 15: Unless pc16 changed things, I am fairly sure this doesn't account for PP Up which is stored as the upper 2 bits. 

Yes, using PP Up on Trump Card is a terrible idea, but I can't control what players do!

engine/battle/move_effects/yawn.asm
```diff
@@ -0,0 +1,28 @@
+BattleCommand_yawn:
+	ld a, [wAttackMissed]
```
FredrIQ, line 2: Everything (except for Uproar) that protects against sleep will also protect against drowsiness. This only checks for an existing status condition (and a miss).

engine/battle/endturn.asm
```diff
@@ -1107,3 +1169,93 @@ HandleRoost:
...
+	ld hl, wEnemyYawnMagnetRiseCount
+.got_yawn
+	ld a, [hl]
+	sub a, $10
+	ret c
+	ld [hl], a
+	sub a, $10
+	ret nc
+	call SwitchTurn
+	farcall BattleCommand_sleep.SleepTarget
```
FredrIQ: Never call local labels like this. If you need to access the routine, make the .label into a global label.

engine/battle/endturn.asm
```diff
@@ -1107,3 +1169,93 @@ HandleRoost:
...
+	call SwitchTurn
+
+.do_it
+	call GetTurnAndPlacePrefix
+	ld hl, wPlayerThroatChopEmbargoCount
+	jr z, .got_throat_chop
+	ld hl, wEnemyThroatChopEmbargoCount
+.got_throat_chop
+	ld a, [hl]
+	sub a, $10
```
FredrIQ: sub $10

engine/battle/endturn.asm
```diff
@@ -1107,3 +1169,93 @@ HandleRoost:
...
+	call SwitchTurn
+
+.do_it
+	call GetTurnAndPlacePrefix
+	ld hl, wPlayerYawnMagnetRiseCount
+	jr z, .got_yawn
+	ld hl, wEnemyYawnMagnetRiseCount
+.got_yawn
+	ld a, [hl]
+	sub a, $10
```
FredrIQ: sub $10 (same for the one right below)

engine/battle/core.asm
```diff
@@ -5670,6 +5741,72 @@ CheckUsableMove:
...
 
+.CheckTaunt:
+	ldh a, [hBattleTurn]
+	and a
+	ld hl, wPlayerTauntCount
+	jr z, .got_taunt_turn
+	ld hl, wEnemyTauntCount
+.got_taunt_turn
+	ld a, [hl]
+	and a, $F0
```
FredrIQ: and $f0

engine/battle/endturn.asm
```diff
@@ -1107,3 +1169,93 @@ HandleRoost:
...
+	call SwitchTurn
+
+.do_it
+	call GetTurnAndPlacePrefix
+	ld hl, wPlayerThroatChopEmbargoCount
+	jr z, .got_embargo
+	ld hl, wEnemyThroatChopEmbargoCount
+.got_embargo
+	ld a, [hl]
+	and a, $0F
```
FredrIQ: and $f

engine/battle/move_effects/magnet_rise.asm
```diff
@@ -0,0 +1,23 @@
...
+	jr nz, .failed
+
+	ldh a, [hBattleTurn]
+	and a
+	ld hl, wPlayerYawnMagnetRiseCount
+	jr z, .got
+	ld hl, wEnemyYawnMagnetRiseCount
+.got
+	ld a, [hl]
+	and a, $0F
```
FredrIQ: and $f

engine/battle/move_effects/throat_chop.asm
```diff
@@ -0,0 +1,16 @@
...
+	ret nz
+
+	ldh a, [hBattleTurn]
+	and a
+	ld hl, wEnemyThroatChopEmbargoCount
+	jr z, .got
+	ld hl, wPlayerThroatChopEmbargoCount
+.got
+	ld a, [hl]
+	and a, $0F
```
FredrIQ: and $f, or $20

engine/battle/move_effects/yawn.asm
```diff
@@ -0,0 +1,28 @@
...
+	jr nz, .failed
+
+	ldh a, [hBattleTurn]
+	and a
+	ld hl, wEnemyYawnMagnetRiseCount
+	jr z, .got
+	ld hl, wPlayerYawnMagnetRiseCount
+.got
+	ld a, [hl]
+	and a, $F0
```
FredrIQ: and $f0

engine/battle/move_effects/taunt.asm
```diff
@@ -0,0 +1,32 @@
...
+	and a
+	jr nz, .failed
+
+	ldh a, [hBattleTurn]
+	and a
+	ld hl, wEnemyTauntCount
+	ld a, [wEnemyGoesFirst]
+	jr z, .got
+	ld hl, wPlayerTauntCount
+	xor a, 1
```
FredrIQ: Instead of this rather non-obvious method of checking who goes first, there is a CheckOpponentWentFirst utility function.

Also, if there is a reason you don't use that function: xor 1

engine/battle/move_effects/taunt.asm
```diff
@@ -0,0 +1,32 @@
...
+	ld b, $30
+	jr z, .order_done
+	; they went first
+	ld b, $40
+.order_done
+	ld a, [hl]
+	and $F0
+	jr nz, .failed
+	ld a, [hl]
+	or a, b
```
FredrIQ: or b

engine/battle/move_effects/yawn.asm
```diff
@@ -0,0 +1,28 @@
...
+	jr nz, .failed
+	ld a, [hl]
+	or a, $20
```
FredrIQ: or $20

engine/battle/move_effects/embargo.asm
```diff
@@ -0,0 +1,23 @@
...
+	and a
+	ld hl, wEnemyThroatChopEmbargoCount
+	jr z, .got
+	ld hl, wPlayerThroatChopEmbargoCount
+.got
+	ld a, [hl]
+	and $0F
+	jr nz, .failed
+	ld a, [hl]
+	or a, $05
```
FredrIQ: or $5

engine/battle/move_effects/magnet_rise.asm
```diff
@@ -0,0 +1,23 @@
...
+	jr nz, .failed
+	ld a, [hl]
+	or a, $05
```
FredrIQ: or $5

engine/battle/endturn.asm
vulcandth: Wouldn't we prefer decimal values in most cases?

vulcandth: With the exception of $0f or $f0 for dealing with nibbles?

engine/battle/move_effects/memento.asm
emilyploszaj: Needs to occur after sub check but before the faint for vanilla parity.

engine/battle/move_effects/trump_card.asm
emilyploszaj: That would only be max PP, this only counts current PP and doesn't care about max value, all values 5+ return 40 power

engine/battle/move_effects/stored_power.asm
emilyploszaj: It's a small difference and this gives us slightly better conveyance and ability to change functionality when/if we change how bp multiplication works.

Issue #1218, opened 2025-06-23
Locks up after beating first "Youngster Joey"


iball:
After finally getting underway once you get to the first "Youngster Joey" and beating him after he says his spiel, the entire game locks up.
Doesn't matter if it's a ROM or flashed to a dedicated cart, the game locks up.
This is the latest "not a real release, for testing only" code pulled from the github as of commit ae13c0d.

amadsen90:
I found this happening with Wade and Anthony as well, seems to be limited to the phone number trainers.  Appears to be random since some of these trainers work properly and some do not.

iball:
Issue still exists as of commit bd0bc08.

DrMungo:
I was able to narrow down the issue a bit. The problem seems to be with `RandomPhoneMon` in `engine/overworld/wildmons.asm`. If I comment out the call to this code on line 496 in `engine/events/std_scripts.asm` Joey correctly continues to ask for the phone number and can be saved in the address book. 
I'm going to investigate a bit further, but also am a complete newb to gbc assembly - so not sure how much insight I'm actually gonna be able to produce.

DrMungo:
Taking a look at the `RandomPhoneMon` implementation, I noticed something off in the `.no_moves` part:
```asm
.no_moves
	; bc == size of mon sub-struct
	ld b, 0

        ; b currently holds party size in bytes
	ld a, b
	add l
	ld e, 0
	push hl
```

The second comment says `b` should currently hold the party size, but is set to `0` on the previous line. And indeed, if I comment out the `ld b, 0`, Joey works as expected. I'll do some more testing with other trainers, asking for the player's number, and getting called by Joey, to check if his boasting works now as well.
What I'm still unsure about though, is whether this line is a simple typo and can be removed, or if something else was supposed to go here. Can someone maybe provide some more insight?

amadsen90:
Removed the RandomPhoneMon from AskNumber1M script and all phone npcs give number with no issues.  Will dig deeper into what is wrong with RandomPhoneMon

orafaelfragoso:
Opened #1244 because this problem still happens on master branch

Issue #1197, opened 2025-05-22
Phone number glitch


Gpsboy97:
I was using the latest build of this game. After I beat Youngster Joey at Route 30, he wanted to ask for a phone number. Instead of the phone number text, the game instantly softlocks as soon as the after-battle text ends.

OxygenTheSmolAngel:
I've been getting this issue as well. My game also softlocks when I get a call from a trainer. It doesn't softlock for me when Lyra, Prof. Elm or my Mom calls me, so this is really odd.

CrispyBuns:
I'm having this exact same issue and I'm on an Everdrive on my GBC so I've noticed for phone numbers (besides Mom, Lyra and Elm) and after fighting trainers that give you their number it softlocks.

Yiran001:
I am facing a similar issue. I got the number from Joey, and everything works fine, but as soon as I go to another route he calls me infinitely.

iball:
Same issue I reported in #1218 and still valid as of the latest commit.

iball:
Issue still exists as of commit bd0bc08.

KlestDedja:
Not the same issue, but may be connected:

The _frequency_ of the calls feels pretty high to me. Also, I get called very often when I load a save.

gbcRascal:
@FredrIQ Was curious if this will be fixed any time? Seems eantonijevic identified it might be from your change.

vulcandth:
Resolved with [3b4f404](https://github.com/Rangi42/polishedcrystal/commit/3b4f40440cb183819804cdb1ed45fdeb0ef84113)

PR #1230, opened 2025-07-17
Fix RandomPhoneMon count_mon loop

initd456:
This should fix issues https://github.com/Rangi42/polishedcrystal/issues/1218 and https://github.com/Rangi42/polishedcrystal/issues/1197.

Reverts back some of the changes from https://github.com/Rangi42/polishedcrystal/pull/1141. `bc` was not correctly set, so it was looping infinitely.

### DIFF
--- a/engine/overworld/wildmons.asm
+++ b/engine/overworld/wildmons.asm
@@ -1095,18 +1095,18 @@ RandomPhoneMon:
 	inc c
 	inc c
 .no_moves
-	; bc == size of mon sub-struct
-	ld b, 0
+	; bc = size of mon struct
+	ld b, c
+	xor c
 
-	; b currently holds party size in bytes
-	ld a, b
-	add l
 	ld e, 0
 	push hl
 .count_mon
 	inc e
 	add hl, bc
-	cp l
+	ld a, [wTrainerGroupBank]
+	call GetFarByte
+	cp $ff
 	jr nz, .count_mon
 	pop hl
 


FredrIQ:
Have you tested this code? Why are you checking for $FF? The terminator is gone, instead the first byte of a party struct is the amount of bytes for the struct. This was to allow the party data to contain 255.

See data/trainers/macros.asm, specifically end_trainer, to see how the trainer party struct is generated.

ghost:
> Have you tested this code? Why are you checking for $FF? The terminator is gone, instead the first byte of a party struct is the amount of bytes for the struct. This was to allow the party data to contain 255.
> 
> See data/trainers/macros.asm, specifically end_trainer, to see how the trainer party struct is generated.

Both commits were tested with youngster joey and hiker anthony. Granted I'm still new to rgbds, so the fix might not be the correct choice.

SunsetteSkye:
> Both commits were tested with youngster joey and hiker anthony. Granted I'm still new to rgbds, so the fix might not be the correct choice.

In testing, this does prevent the locked loop, but I'm getting complete gibberish in the place of monster names -- first letter combinations that made no sense but seemed to be a straight string of characters from the naming interface, now random sprite selections (that I think are from the Pokemon Center north wall). The calls I've received for a rematch or an item have been false positives as well.

ghost:
> > Both commits were tested with youngster joey and hiker anthony. Granted I'm still new to rgbds, so the fix might not be the correct choice.
> 
> In testing, this does prevent the locked loop, but I'm getting complete gibberish in the place of monster names -- first letter combinations that made no sense but seemed to be a straight string of characters from the naming interface, now random sprite selections (that I think are from the Pokemon Center north wall). The calls I've received for a rematch or an item have been false positives as well.

@SunsetteSkye  Are you testing on a new game or existing game and which commit? It worked fine for me on a new game and saved game with the latest commit; see below.

https://github.com/user-attachments/assets/a856955c-ffcf-4a18-a376-0bfd2d9ec692

https://github.com/user-attachments/assets/d580122a-b9a7-4160-aef8-182be1db9543

SunsetteSkye:
Sometime around Goldenrod, phone calls became completely normal again. No further changes made.

ghost:
@SunsetteSkye It would be helpful if you can mention which trainers.

SunsetteSkye:
> @SunsetteSkye It would be helpful if you can mention which trainers.

I think that was the right call. It may have appeared to get better at Goldenrod because I had a bigger phone book.

Calls are random, but I've noticed it four times with Wade since I started noting which trainer was linked. So far haven't seen it on anyone else. 

image

ghost:
Closing as I dont have time to debug this further to fix. Maybe someone can.

PR #1231, opened 2025-07-19
Remove cp macros and inline comparisons

vulcandth:
Sylvie didn't want the macros' and I agree with that now. They are removed. :)


emilyploszaj (changes requested):
Reasonable implementation, I think I'd prefer if style was consistent, `.cphl_fly` instead of `.cphl_fly_ai`. Also, what do you think about including asserts for non zero high byte comparison?

PR #1236, opened 2025-07-31
Use features new to RGBDS 0.9.4

Rangi42:
This uses `'character'` constants for numeric values instead of single-character `"strings"`. It also passes `-Weverything` to all four programs.

The ROM hash is unchanged.


vulcandth (approved):
LGTM!

PR #1244, opened 2025-08-27
Fix phone NPC freeze after trainer battles

orafaelfragoso:
## Problem
The game would freeze after defeating Youngster Joey (and potentially other trainers) when the phone number request dialog appeared. This was caused by a critical bug in the `RandomPhoneMon` function in `engine/overworld/wildmons.asm`.

### Root Cause
The `RandomPhoneMon` function was responsible for selecting a random Pokémon from a trainer's party to display in phone number request messages. However, the trainer party counting logic was completely broken:

1. **Faulty party size calculation**: The code attempted to count Pokémon in a trainer's party but used incorrect logic
2. **Broken loop termination**: The counting loop compared against an uninitialized value (`a = 0`), causing it to either loop indefinitely or access invalid memory
3. **Missing error handling**: No safeguards existed for edge cases or malformed trainer data

### Technical Details
The problematic code was:
```asm
; bc == size of mon sub-struct
ld b, 0

; b currently holds party size in bytes
ld a, b        ; a = 0 (b was set to 0)
add l          ; This doesn't change hl  
ld e, 0
.count_mon
inc e
add hl, bc
cp l           ; Comparing 0 with l - makes no sense!
jr nz, .count_mon
```

This would cause the function to malfunction when trying to count trainer party Pokémon, leading to memory access violations or infinite loops.

### Solution
Completely rewrote the party counting logic in `RandomPhoneMon`:

1. **Proper party counting**: Now correctly reads trainer data byte-by-byte, counting valid Pokémon entries until reaching the end marker
2. **Bounds validation**: Added checks to ensure trainers have at least 1 Pokémon  
3. **Error handling**: Implemented fallback to PIDGEY if data parsing fails
4. **Safe memory access**: Ensured all pointer operations stay within valid bounds

### Files Changed
- `engine/overworld/wildmons.asm`: Fixed `RandomPhoneMon` function party counting logic

### Testing
The fix resolves the freeze that occurred after defeating Youngster Joey and requesting his phone number. The phone system now correctly displays a random Pokémon from the trainer's party in the phone number request dialog.

### Related Issues
- Fixes issue #1218
- Also fixes potential freezes with other phone-enabled trainers

This fix ensures the phone NPC system works reliably for all trainer battles without causing game freezes.

### DIFF
--- a/engine/overworld/wildmons.asm
+++ b/engine/overworld/wildmons.asm
@@ -928,50 +928,66 @@ RandomPhoneRareWildMon:
 
 .GetGrassmon:
 	push hl
-	ld bc, 5 + 4 * 3 ; Location of the level of the 5th wild Pokemon in that map
+	; Skip map ID (2 bytes) and encounter rates (3 bytes)
+	ld bc, 5
 	add hl, bc
+
+	; Get current time of day and skip to that section
 	call GetTimeOfDayNotEve
 	ld bc, NUM_GRASSMON * 3
 	rst AddNTimes
+
+	; Select one of the last 3 entries (rarest Pokémon)
 	ld a, 3
 	call RandomRange
 	ld c, a
-	ld b, $0
+	ld b, 0
+	ld de, 4 * 3 ; Skip first 4 entries to get to entries 4, 5, 6
+	add hl, de
+	add hl, bc ; Add random offset (0-2 entries)
 	add hl, bc
 	add hl, bc
-	add hl, bc
-; We now have the pointer to one of the last (rarest) three wild Pokemon found in that area.
+
+	; Skip level byte, read species and form
 	inc hl
-	ld a, [hli] ; Contains the species index of this rare Pokemon
+	ld a, [hli] ; Species
 	ld c, a
-	ld a, [hl] ; Contains the form (including extspecies)
+	ld a, [hl] ; Form
 	ld b, a
 	ld [wCurForm], a
 	pop hl
-	ld de, 5 + 0 * 3
+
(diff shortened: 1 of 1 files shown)


Basetezotar:
According to developer comments in the Discord server, this bug has been left in the prereleased version intentionally, to discourage players from using it while it is still under development, because it will not be save-compatible with the final version. It will be fixed when 3.2 is ready for release.

FredrIQ:
Fix looks good for the most part though, I'll merge this when 3.2 is to be released.

That being said, I feel like instead of using Pidgey as a failsafe, the Crash handler should be used.

orafaelfragoso:
> Fix looks good for the most part though, I'll merge this when 3.2 is to be released.
> 
> That being said, I feel like instead of using Pidgey as a failsafe, the Crash handler should be used.

That sounds good. I will make the adjustment.

The 3.2 release is so good, I wanted to play it but got stuck, so I fixed it for myself and shared the results.

gbcRascal:
@FredrIQ When is 3.2 release? If it's a long while, is there a dev branch you can merge this into? I would like to use the latest working version for my hack mods

melthelesbian:
This works around the freeze, but still results in garbage data in the actual calls, so this fix isn't quite ready for merging. (tested on retroarch's sameboy core)
Polished Crystal (v3 2 0-dev phone-numbers-trainers-lock)-250920-151542

vulcandth:
Closing this as it was superseded. FIQ may want to re-look at it later, so please wait a while before deleting the branch. (We may want to reopen it)

PR #1245, opened 2025-09-07
Moves Batch

aubsgp:
Added: Wake-Up Slap, Smelling Salts, Strange Steam, Mud Sport, Water Sport, Magic Room, and Wonder Room. 

Some other miscellaneous edits, such as: 
- Combined Trick Room and Gravity into one byte
- Converted some constants into static variables (bitmasks mainly)
- Added a general-purpose DecrementCounter to save space on end-of-turn nybble decrements (with an associated message if they hit 0)

### DIFF
--- a/engine/math/math.asm
+++ b/engine/math/math.asm
@@ -234,24 +234,3 @@ Divide16::
 .division_by_zero
 	ld a, ERR_DIV_ZERO
 	jmp Crash
-
-DivideBy12::
-	srl a
-DivideBy6::
-	srl a
-; Divides a by 3 by multiplying by .0101011, or about 0.336.
-DivideBy3:: 
-	push bc
-	ld b, a
-	srl a
-	add b 
-	srl a
-	srl a
-	add b
-	srl a
-	srl a
-	add b
-	srl a
-	srl a
-	pop bc
-	ret


data/moves/descriptions.asm
```diff
@@ -1970,3 +1977,38 @@ ViseGripDescription:
...
+
+MagicRoomDescription:
+	text "Negates held items"
+	next "for five turns."
+	done
+
+WonderRoomDescription:
+	text "Temporarily swaps"
+	next "Def with Sp.Def."
+	done
```
emilyploszaj: Terminating newline here

data/text/battle.asm
```diff
@@ -649,6 +649,11 @@ WokeUpText:
 	line "woke up!"
 	prompt
 
+WokeUpOpponentText:
```
emilyploszaj: Is this necessary? Isn't there already text for waking up?

data/text/battle.asm
```diff
@@ -1262,6 +1267,52 @@ TrickRoomEndedText:
 	line "normal!"
 	prompt
 
+MagicRoomText:
+	text "It created a"
```
emilyploszaj, line 1266: Unless this is the canon text, prefer `<USER> created...`

data/text/battle.asm
```diff
@@ -1369,6 +1420,12 @@ AlreadyParalyzedText:
 	line "already paralyzed!"
 	prompt
 
+HealParalysisOpponentText:
```
emilyploszaj: Is this canon text?

engine/battle/effect_commands.asm
```diff
@@ -3836,12 +3835,38 @@ BattleCommand_damagestats:
 	ld d, a
 	ret z
 
+; Check for wonder room. If present, we store $ff in e to xor with later.
+	ld e, $00
+	ld a, [wMagicWonderRoom]
+	and FIELD_WONDER_ROOM
+	jr z, .which_attack
+	ld e, $ff
```
emilyploszaj: `dec e`

engine/battle/effect_commands.asm
```diff
@@ -3836,12 +3835,38 @@ BattleCommand_damagestats:
...
+
+.which_attack
```
emilyploszaj: `.got_wonder_room`

engine/battle/effect_commands.asm
```diff
@@ -3836,12 +3835,38 @@ BattleCommand_damagestats:
...
 	ld a, BATTLE_VARS_MOVE_CATEGORY
 	call GetBattleVar
 	cp SPECIAL
-	jr nc, .special
-
-.physical
+	jr nc, .special_attack
+.physical_attack
+; If the attack is physical, we make a note of it in b by co-opting the reflect mask.
```
emilyploszaj: For this whole section, commenting instructions should be done indented on its own line. Comments can exist after instructions if highly logical but prefer line above.

engine/battle/effect_commands.asm
```diff
@@ -3836,12 +3835,38 @@ BattleCommand_damagestats:
...
+	ld b, SCREENS_REFLECT
+	xor a ; Default a to $00, which is the value signalling "physical defense" later.
+	jr .which_defense
+.special_attack
+; If the attack is special, we make a note of it in b by coopting the light screen mask
+	ld b, SCREENS_LIGHT_SCREEN
+	ld a, BATTLE_VARS_MOVE_EFFECT
+	call GetBattleVar
+	cp EFFECT_PSYSTRIKE
+	ld a, $ff ; By default, set a to $ff = special
```
emilyploszaj: `xor a`, `jr z, .which_defense`, `inc a`

engine/battle/effect_commands.asm
```diff
@@ -3851,27 +3876,8 @@ BattleCommand_damagestats:
 	call HailDefenseBoost
 	call DittoMetalPowder
```
emilyploszaj: Do these boosts now only apply for physical defense?

engine/battle/effect_commands.asm
```diff
@@ -4185,7 +4220,7 @@ ConfusedDamageCalc:
 	call DamagePass1
 	call DamagePass2
 	push bc
-	; This way we ignore Unnerve
+	; This way we ignore Unnerve <- (do they mean Unaware?)
```
emilyploszaj: Yes, update comment

engine/battle/effect_commands.asm
```diff
@@ -7004,3 +7043,31 @@ GetOpponentWeight::
 	srl h
 	rr l
 	ret
+
+BattleCommand_wakeopponent:
```
emilyploszaj: Put these in `wake_up_slap.asm` and `smelling_salts.asm`, also, put in battle commands 2 if possible.

engine/battle/effect_commands.asm
```diff
@@ -7004,3 +7043,31 @@ GetOpponentWeight::
...
+	call GetBattleVarAddr
+	ld b, a
+	and SLP_MASK
+	ret z
+	ld a, b
+	and ~SLP_MASK
+	ld [hl], a
+	call RefreshBattleHuds
+	ld hl, WokeUpOpponentText
+	call StdBattleTextbox
```
emilyploszaj: Jump instead of call ret

engine/battle/effect_commands.asm
```diff
@@ -7004,3 +7043,31 @@ GetOpponentWeight::
...
+	call GetBattleVarAddr
+	ld b, a
+	and (1 << PAR)
+	ret z
+	ld a, b
+	and ~(1 << PAR)
+	ld [hl], a
+	call RefreshBattleHuds
+	ld hl, HealParalysisOpponentText
+	call StdBattleTextbox
```
emilyploszaj: Jump instead of call ret

engine/battle/move_effects/mud_sport.asm
```diff
@@ -0,0 +1,16 @@
...
+    ld hl, wFieldSports
+    ld a, [hl]
+    and FIELD_MUD_SPORT
+    jr nz, .failed
+    ld a, [hl]
+    add $50
+    ld [hl], a
+    ld hl, MudSportText
+    jmp StdBattleTextbox
+    ret
```
emilyploszaj: Redundant ret

engine/battle/move_effects/mud_sport.asm
```diff
@@ -0,0 +1,16 @@
...
+
+.failed
+    farcall AnimateFailedMove
+	farcall TryPrintButItFailed
```
emilyploszaj: farjump instead of farcall ret

ram/wram0.asm
```diff
@@ -602,17 +602,27 @@ wWhichMonFaintedFirst:: db
...
-
 wBattleLowHealthAlarm:: db
 
 wPlayerSemiInvulnerableType:: db
 wEnemySemiInvulnerableType:: db
 
-wFieldEffects::
+wFieldSports::
+; bit
+; 4-7 water
```
emilyploszaj: Say water sport and mud sport here instead of just water and mud

ram/wram0.asm
```diff
@@ -1058,7 +1068,7 @@ wSummaryMoveSwap:: db
 
 ; Used to align window buffer for DMA copying
 ; Feel free to use or move data, an assert will fail if the memory becomes misaligned
-ds 13
+ds 12
```
emilyploszaj: Take this byte away from the unused space in the battle section instead of here.

emilyploszaj (commented):
Lots of files are missing trailing newline. Left comments on specific nitpicks, but there were no major logical problems.

data/text/battle.asm
aubsgp: While there is, it's for the Pokemon whose turn it currently is. I suppose it would be reasonable to switch turns, play the wake-up message, then switch back? Would that be preferable?

data/text/battle.asm
aubsgp: This is the canon text, yeah, or close to it.

"\<User\> used Wonder Room!"
"It created a bizarre area in which Pokemon's held items lose their effect!"

data/text/battle.asm
aubsgp: The canon text is:

"\<Target\> was healed/cured of paralysis!"

I can switch to that, sorry for not being precise.

engine/battle/effect_commands.asm
aubsgp: Yes, Hail and Metal Powder only boost Defense.

engine/battle/effect_commands.asm
aubsgp: I can't do this because xor a would mess with the z flag.

aubsgp: Also, did you mean dec a? I'm not sure I understand inc a, unless a is already $ff and we want to turn it into $00?

engine/battle/effect_commands.asm
aubsgp: Do you mean effect_commands_2? Currently both wake_up_slap and smelling_salts are INCLUDEd in that file, if so.

constants/battle_constants.asm
```diff
@@ -276,11 +276,24 @@ DEF TEAM_LUCKY_CHANT EQU %00000111
 DEF TEAM_TAILWIND    EQU %00111000
 DEF TEAM_WISH        EQU %11000000
 
+; wPlayerThroatChopEmbargoCount or wEnemyThroatChopEmbargoCount bit masks
+DEF EMBARGO_MASK     EQU %00001111
```
emilyploszaj: everything else like this is named with a prefix and without the `_MASK`, maybe this could be as well?

engine/math/math.asm
```diff
@@ -234,3 +234,24 @@ Divide16::
...
+	srl a
+	srl a
+	add b
+	srl a
+	srl a
+	add b
+	srl a
+	srl a
+	pop bc
+	ret
```
emilyploszaj: Still missing a terminating newline

polishedcrystal-debug-3.2.0.ss1
emilyploszaj, line 1: Don't commit savestate files

engine/battle/effect_commands.asm
```diff
@@ -3835,93 +3834,144 @@ BattleCommand_damagestats:
...
 	ld a, MON_SAT
 	call TrueUserPartyAttr
-.sat_ok
+
+.screens
+; Double defense if screens are active.
 	call GetOpponentActiveScreens
-	and SCREENS_LIGHT_SCREEN
-	jr z, .lightball
+	and e
```
emilyploszaj, line 3932: is this supposed to be `a + e`?

aubsgp: No, e contains the physical/special status of the move, but stored by co-opting the corresponding screen.

data/text/battle.asm
FredrIQ: That is standard practice in Polished, so I'd prefer that, yes (The only argument I could see as potential is if doing so would be less space-efficient... but it isn't).

FredrIQ:
Thanks for the PR!

I will take a closer look at the code shortly. While it looks fine logically, there are some behaviour I am unsure about whether it matches mainline or not (particularly around Wonder Room) which I'll have to look over.

One thing I'm wondering is: what made you choose these particular moves to implement? They are for the most part non-controversial so I see little issue in adding them. I find Mud/Water Sport to be utterly pointless moves (challenge: come up with a case where Mud Sport is legitimately useful), but I'm not outright against adding them. But there are moves in mainline that we don't really want to add, such as Stealth Rock, which is why I'm asking how you decided what moves to add.

aubsgp:
Yeah, Wonder Room is really weird, and it took me a while to run through all the behavior on Showdown and the like, wouldn't be surprised if I missed something. Gotta say, I was not expecting to learn that Download ignores it!

As for how I chose these moves -- I was sort of trying to gain experience, mostly. There are some fun niche uses of Mud/Water Sport in harder romhacks, where the player doesn't get tools like Rain Dance, but they are mostly useless moves even there...I just like them, I guess. :)

(I did also double-check with Emi that you would in fact want these moves; she's mentioned the stealth rock thing to me before.)

Thanks for taking the time to look this over!

polishedcrystal-debug-3.2.0.ss1
rofl0r, line 1: checked in by mistake ?

aubsgp: Yep, sorry.

aubsgp:
Mud Sport, Water Sport, and Smelling Salts have been all removed, as requested. Let me know if you'd like any additional changes.

engine/battle/effect_commands.asm
```diff
@@ -3966,95 +3834,121 @@ BattleCommand_damagestats:
 	and a
 	ld d, a
 	ret z
+	; We won't need base power for a while, so let's push it temporarily to make room.
+	push de
+
+; Check for wonder room. If present, we store $ff in e to xor with later.
```
FredrIQ: You are making Wonder Room more complicated than necessary, possibly due to a lack of complete understanding of what the move actually does.

Basically, Wonder Room does 2 things, only 1 which is actually intentional. The main part is that it swaps the Defense and Special Defense stat *before* modifiers such as stat changes or items. Body Slam for example will use the foe's Sp. Def stat, but still factor in Reflect, +Def stat changes, and will not care about Assault Vest.

Wonder Room also has an additional effect, which is actually a bug: Download will ignore Wonder Room for the purpose of determining which stat to use, but *will* use the opposite stat change. Specifically, it will compare the foe's Defense with Sp.Def. stat changes, and Sp.Def with Defense stat changes. Body Press does something similar (it calculates Defense with Sp.Def stat changes), but we don't have that move in the game. Since Polished doesn't factor in stat changes for Download in the first place (which is in error on my part), you don't really need to worry about this, I'll deal with it separately.

While you might have a better idea, I would solve this by simply making dedicated routines for "GetOpponentDefenseStat" / "GetOpponentSpDefenseStat" which stores the correct, Wonder Room-switched, stat in bc, and leave the function unchanged otherwise.

PR #1247, opened 2025-09-07
Define routine documentation format

emilyploszaj:
This PR proposes a format for rigorous routine documentation and applies it to a bunch of high traffic routines. Let me know any thoughts you have on the formatting, it's not set in stone.

Additionally, a `fallthrough` macro has been added. All it does is validate that the fallthrough hasn't had code inserted, slightly more reasonable than a fallthrough comment.

Once this has been discussed and merged, I'll be applying more documentation across high traffic routines in the codebase, particularly around 10bit move expansion. If it ends up being useful, we can think about porting it upstream to pokecrystal.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,68 @@
+# Routine Documentation
+
+When commenting routines be sure to follow this style guide for consistency.
+Many different styles of comments exist from various points of time but, this is considered up to date.
+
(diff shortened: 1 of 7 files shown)


vulcandth:
Most of our routine comments are _under_ the routine label. Is there a precedent for IDE's to read it above the label?

PR #1259, opened 2025-10-02
DayCare: use shorter text for "want egg" query after first egg

rofl0r:
if you're breeding lots of eggs, clicking through the same monologue consisting of several screens over and over gets tiring soon.

instead, we reuse the existing flag for the first bred egg to use a shorter text from then on, similar to how the nurse talks less after becoming the champion.


data/text/common.asm
```diff
@@ -1634,6 +1634,15 @@ _NotYetText::
 	text "Not yet…"
 	done
 
+SECTION "_FoundAnEggAgainText", ROMX
+_FoundAnEggAgainText::
+	text "Hello <PLAYER>!"
+	line "Your #mon had"
+
+	para "another egg."
+	line "You want it?"
```
Rangi42: text "Ah, it's you!"

	para "Your #mon had"
	line "another Egg!"

	para "You want it?"

engine/events/daycare.asm
```diff
@@ -395,8 +395,18 @@ Special_DayCareManOutside:
 	text_far _NotYetText
 	text_end
 
-.AskGiveEgg:
+.SelectAskGiveEggText:
```
Rangi42: This can be inlined, as it's only called once.

engine/events/daycare.asm
```diff
@@ -396,7 +396,14 @@ Special_DayCareManOutside:
 	text_end
 
 .AskGiveEgg:
+	eventflagcheck EVENT_BRED_AN_EGG
+	jr z, .not_bred_yet
+	; if user already bred an egg, use shorter monologue.
+	ld hl, .IntroAgainText
+	jr .show_question
+.not_bred_yet
```
Rangi42: https://github.com/pret/pokecrystal/wiki/Optimizing-assembly-code
```suggestion
	; if user already bred an egg, use shorter monologue.
	eventflagcheck EVENT_BRED_AN_EGG
	ld hl, .IntroAgainText
	jr nz, .show_question
```

Rangi42:
Thank you!

rofl0r:
nice, thanks for taking the time to look at it this quickly, and also the link to the optimization tutorial, wish i would have found it earlier :)

PR #1263, opened 2025-10-11
Fixed two bugs with how the high bits of 10 bit moves were handled during PC storage.

aubsgp:
- On encoding, the byte containing the high bits of the moves was rotated an extra time, which gave each move the high bits of the move after it (and killed the high bits of the last move).
- On decoding, an unnecessary mask was performed with an incorrect bitmask (%00111111 instead of %11111100), killing the high bits of the first move. This mask has been removed, as it seems to serve no discernible purpose.


vulcandth:
Ah yeah. I see the mistake I made. Thanks! Looks good to me.

PR #1275, opened 2025-10-25
QOL: turn player's head down after arriving in magnet train station

rofl0r:
when clicking through the "hope to see you again" text after arrival, one button press too much causes one to trigger a new departure dialog which one has to be careful about in order not to board the train again.

therefore, turn the player's direction down after the monologue so pressing A once too often can't result in this situation - similar to how after healing mons in the pokemon center the player faces away from the nurse.


JanJastrow:
I'm using this issue because I noticed a similar thing a few days ago:

When using the bed on S.S. Aqua the player also does not turn around after sleeping – so you can quickly loop through another "night" by accident.
I don't know if that's the case with other beds, too.

vulcandth:
Resolved in 6e5d21da1aa7738c7bb106217f66aeb6e56b9b33

PR #1278, opened 2025-10-26
Fix scientist side sprite

rjd1922:
Changed the scientist's static side sprite to include the temples of his glasses like in Gen I.


Rangi42:
Thanks!

PR #1293, opened 2025-11-08
Overhaul Mold Breaker

FredrIQ:
This overhauls Mold Breaker, fixing several smaller inconsistencies in how it worked. This also allowed us to remove various special cases in the code that had to work around the way Mold Breaker currently worked.

From an accuracy standpoint, assuming there's no outright bugs in the code, this should fully replicate the way Mold Breaker behaves in current vanilla, following the spec at https://wiki.xn--rckteqa2e.com/wiki/%E3%81%8B%E3%81%9F%E3%82%84%E3%81%B6%E3%82%8A#%E7%89%B9%E6%80%A7%E3%81%AE%E4%BB%95%E6%A7%98 . Some examples of what this fixes:

* Item-induced stat changes as a result of an opponent move now have their abilities ignored when applicable for Mold Breaker.
* Several issues with hazards + Mold Breaker phazing was addressed. Previously, only MB Roar into Levitate hazards was properly handled
* Mold Breaker no longer applies to Future Sight.

I posted this as a PR in case we have any future 3.2 bugfix releases, because I feel that the underlying code changes were too large to justify going into a bugfix release. I'm also asking @emilyploszaj for a code review for overall code quality, and for potential improvements to it, since Emi spends plenty of time working on battle engine code and I don't wish to make their life harder.

### DIFF
--- a/constants/battle_constants.asm
+++ b/constants/battle_constants.asm
@@ -278,10 +278,10 @@ DEF GUARD_MIST      EQU %11110000
 
 ; wMoveState
 	const_def
-	const MOVESTATE_PHYSICAL_F
-	const MOVESTATE_SPECIAL_F
-	const MOVESTATE_IGNOREABIL_F
-	const MOVESTATE_ENDED_F
+	const MOVESTATE_PHYSICAL_F ; move is physical (needed for counter)
+	const MOVESTATE_SPECIAL_F ; move is special (needed for mirror coat)
+	const MOVESTATE_IGNOREABIL_F ; move ignores Abilities (Mold Breaker)
+	const MOVESTATE_ENDED_F ; CheckTurn failure flag
 	const MOVESTATE_OPP_PHYSICAL_F
 	const MOVESTATE_OPP_SPECIAL_F
 	const MOVESTATE_OPP_IGNOREABIL_F
--- a/engine/battle/abilities.asm
+++ b/engine/battle/abilities.asm
@@ -1985,7 +1985,8 @@ _GetTrueUserIgnorableAbility::
 	call GetTrueUserAbility
 	call StackCallOpponentTurn
 _GetIgnorableAbility:
-; Returns the user/target's ability unless the other side is ignoring Abilities.
+; Check if user's Mold Breaker ignores opponent's ability a.
+; Returns the opponent's ability unless the other side is ignoring Abilities.
 	push de
 	push bc
 	ld b, a
--- a/home/battle.asm
+++ b/home/battle.asm
@@ -504,7 +504,7 @@ GetOpponentAbility::
 	ret
 
 GetTrueUserIgnorableAbility::
-; Returns an opponent's ability unless Mold Breaker
+; Returns an user's ability unless Mold Breaker
 ; will suppress it. Preserves bc/de/hl. Note that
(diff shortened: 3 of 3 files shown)


engine/battle/abilities.asm
```diff
@@ -1971,17 +1978,26 @@ RegeneratorAbility:
...
-	call GetTrueUserAbility
-	cp MOLD_BREAKER
-	ld a, b
+	ldh a, [hBattleTurn]
+	and a
+	ld a, [wMoveState]
+	jr z, .got_movestate
+	swap a
+.got_movestate
+	and MOVESTATE_IGNOREABIL
```
emilyploszaj, line 1999: Is this jump inverted? I'd expect mold break to be active when MOVESTATE_IGNOREABIL is set

home/battle.asm
```diff
@@ -503,10 +503,16 @@ GetOpponentAbility::
 	pop af
 	ret
 
-GetOpponentAbilityAfterMoldBreaker::
+GetTrueUserIgnorableAbility::
 ; Returns an opponent's ability unless Mold Breaker
```
emilyploszaj: opponent -> user

constants/battle_constants.asm
```diff
@@ -276,6 +276,28 @@ DEF GUARD_MIST      EQU %11110000
 	const ATKFAIL_ACCMISS   ; missed from accuracy
 	const ATKFAIL_CUSTOM    ; custom message
 
+; wMoveState
+	const_def
```
emilyploszaj, line 280: Some of these are pretty self explanatory, but I would like some documentation here, it's not clear what "ended" means and what it'd be used for, for example.

emilyploszaj (commented):
"Ignorance" is a bit of a funny word to use with english context, but I wouldn't call it wrong. Other than the comments, this looks cleaner. Just curious, not a blocker, have you run dream eater tests on it?

engine/battle/abilities.asm
FredrIQ: _GetIgnorableAbility checks whether the opponent's ability is being ignored by user's Mold Breaker. Thus, if hBattleTurn==0 (player's turn), the IGNOREABIL flag is in the lower nibble.

With that said, I can see how this logic could be rather confusing, but I feel that checking for user's ability being ignored by opponent's Mold Breaker could provide equally, if not more, confusing. Any suggestions for how I can make this more straightforward code-wise, or should I just add more comments to clarify this fact?

constants/battle_constants.asm
FredrIQ: ENDED is used for EndTurn, which is the way CheckTurn "returns" the result. Likely from vanilla (wMoveState is Polished-specific, but incorporated vanilla's wTurnEnded as part of it). I could probably get rid of it alltogether in favour of returning carry or similar, but that feels out of scope for this PR, so I'll just add comments for it alongside the other flags for now.

FredrIQ:
> "Ignorance" is a bit of a funny word to use with english context, but I wouldn't call it wrong. Other than the comments, this looks cleaner. Just curious, not a blocker, have you run dream eater tests on it?

I have not run Dream Eater on it. Doesn't Dream Eater rely on code that isn't in master? I wanted to get back to writing Dream Eater tests when the project was more straight-forward to manage.

PR #1295, opened 2025-11-13
Use RGBDS 1.0.0 features

vulcandth:
- [x] Use single-quoted character literals or CHARVAL when you need a character's numeric value; using strings as numbers is deprecated
- [x] Use ? to silence a node in warning/error location backtraces (@Rangi42 May want to review this)
- [x] Use === and !== instead of STRCMP, and ++ instead of STRCAT
- [x] Use __SCOPE__ instead of checking for .. and . separately
- [x] Use #string instead of "{string}"
- [x] All programs (rgbasm, rgblink, rgbfix, rgbgfx) support -W warnings
- [x] rgbgfx --colors dmg is short for rgbgfx --colors dmg=e4

I didn't really touch the `Makefile`. I'll let you decide what you want to do with that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,42 +27,43 @@ Q :=
 
 POCKET_LOGO = gfx/logo/pocket.bin
 
-RGBASM_FLAGS     = -Weverything -Wtruncation=1 -E -Q8 -P includes.asm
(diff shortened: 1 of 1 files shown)


Rangi42 (approved):
LGTM! I'll wait to let @vulcandth or @FredrIQ merge this once we definitely all have RGBDS 1.0.0.

PR #1299, opened 2025-11-28
Rework Dyn Obj Pals + OW Fading

### LINKED ISSUE #1150: Returning from Stormy Beach
Returning from Stormy Beach to upper area, your character + some NPCs stays faded.

Video soon

vulcandth:
Rework the Dynamic pal system such that we:

- Keep track of the previous palette states before fading (ex. previous map when going through a map connection)
- When an object spawns while fading, we use the previous state to load what it would have looked like on the previous map, current state to load what it will look like on the current map as our target, and then Run a catchup routine to sync the palette with the rest of the palettes as if it were fading the whole time.

resolves #1150

### DIFF
--- a/engine/gfx/fade.asm
+++ b/engine/gfx/fade.asm
@@ -190,10 +190,10 @@ FadePalettesStep:
 	ld a, [wPalFadeMode]
 	and PALFADE_WHICH
 	ld hl, wBGPals2
-	ld d, PAL_COLORS * 16 ; colors * number of palettes
+	ld d, PAL_COLORS * 16 ; bg + ob pals
 	jr z, .got_count
 	dec a
-	ld d, PAL_COLORS * 8 ; colors * number of palettes
+	ld d, PAL_COLORS * 8 ; ; ob pals only
 	jr z, .got_count
 	ld hl, wOBPals2
 .got_count
@@ -301,9 +301,9 @@ FadeColorGetGreen:
 	rrca
 ; fallthrough
 FadeColorGetBlue:
+	and COLOR_BLUE
 	rrca
 	rrca
-	and COLOR_CH_MAX
 	ret
 
 FadeColorStep:


engine/gfx/sprite_palettes.asm
```diff
@@ -190,6 +256,97 @@ ApplyOBPals:
...
+	ld l, e
+	ldh a, [rWBK]
+	push af
+	ld a, BANK(wOBPals1)
+	ldh [rWBK], a
+	; we are not loading an official palette,
+	; so this tells dynamic pals to not associate this
+	; palette with a sprite.
+	ld a, $ff
+if !DEF(MONOCHROME)
```
Rangi42:
```suggestion
if !DEF(MONOCHROME)
	ld a, $ff
```

engine/gfx/fade.asm
```diff
@@ -190,10 +190,10 @@ FadePalettesStep:
 	ld a, [wPalFadeMode]
 	and PALFADE_WHICH
 	ld hl, wBGPals2
-	ld d, 4 * 16 ; colors, palettes
+	ld d, PAL_COLORS * 16 ; colors * number of palettes
```
Rangi42: I think the comment can just be removed, the constant usage is clear enough. Same for the one below. If you think it needs clarification, "`; bg + ob pals`" and below "`; ob pals only`" could be more helpful.

engine/gfx/fade.asm
```diff
@@ -292,20 +292,21 @@ FadePalettesStep:
...
 	rr c
 	ld a, c
 	rrca
-.getBlue:
+; fallthrough
+FadeColorGetBlue:
 	rrca
 	rrca
-	and %00011111
+	and COLOR_CH_MAX
```
Rangi42: If you're going to use `COLOR_RED` and `COLOR_GREEN_HIGH/LOW` above, may as well use `COLOR_BLUE` here for consistency/clarity, so the `and` would come before the two `rrca`s.

PR #1325, opened 2026-01-04
Animate the keyboard in the music player

ariscop:
Could probably look better, some contrast between notes and key colours, but thats outside my skill set


Rangi42:
This makes a surprisingly big difference in being able to follow the music as it's played! I added more contrast between the pressed keys and the notes:

image

PR #1327, opened 2026-01-07
Interrupt cleanup

ariscop:
Cleans up the rIF/rIE code in vblank, a relic of vanilla

Fixes a visual error in music player and  the pokedex clobbering hROMBankBackup, which happened to work previously but is broken with this change

### DIFF
--- a/ram/wram0.asm
+++ b/ram/wram0.asm
@@ -1107,9 +1107,7 @@ wPrinterOpcode:: db
 
 SECTION UNION "Misc 1326", WRAM0
 
-	ds 176
-	assert @ & $FF == 0, "wMPNotes must be 8-bit aligned"
-wMPNotes:: ds 4 * 256
+wMPNotes:: ds 3 * 256
 
 
 SECTION "Video", WRAM0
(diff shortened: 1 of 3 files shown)


ram/wram0.asm
```diff
@@ -1105,6 +1105,13 @@ wGameboyPrinterRAMEnd::
 wPrinterOpcode:: db
 
 
+SECTION UNION "Misc 1326", WRAM0
+
+	ds 176
+	assert @ & $FF == 0, "wMPNotes must be 8-bit aligned"
```
Rangi42: I think this can be just `align 8`, *if* engine/pc/bills_pc_ui.asm adds this:
```diff
 BillsPC_LCDCode:
 LOAD UNION "Misc 1326", WRAM0
+	; TODO: keep in sync with ram/wram0.asm, remove with RGBDS 1.0.2
+	; (see https://github.com/gbdev/rgbds/issues/1881)
+	align 8, -176
 wLCDBillsPC1::
```
And/or swap the order of objects in the Makefile.

ariscop: Not a fan of either solution, i would prefer the assert as its self-contained, but there was enough room in hblank to remove the need for alignment altogether

Rangi42:
Thank you!

PR #1328, opened 2026-01-10
Clean up audio channel rest logic

ariscop:
Replaces ClearChannel with less aggressive per-channel logic. In particular it avoids turning the ch3 dac off constantly


Rangi42:
Nice, thank you! Things sound the same as before. :+1:

PR #1342, opened 2026-01-24
LZ compress map name sign graphics

Rangi42:
The pointer table uses 13 more bytes, but then compressing the graphics saves 360 bytes, for a total of 347 bytes saved.

### DIFF
--- a/engine/events/map_name_sign.asm
+++ b/engine/events/map_name_sign.asm
@@ -221,11 +221,11 @@ LoadMapNameSignGFX:
 	sub l
 	ld h, a
 	ld a, [hli]
-	ld e, a
-	ld d, [hl]
-	ld hl, vTiles0 tile POPUP_MAP_FRAME_START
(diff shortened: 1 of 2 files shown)


rofl0r:
an alternative to trying to squeeze every single byte could be to just remove the RTC functionality and switch to MBC5. imo using the RTC is quite a pain anyway (anyone who has a job will never be able to play at certain day/hour combos) and the replacement that polished offers is superior in every way.

Rangi42:
Yeah, switching to MBC5/MBC3O is inevitable by now, with just over 1% of the ROM still free, but it's fun to see how far we can stretch that last 1%. :P (One whole ROM bank is still empty!)

Also, when we do finally switch to a 4 MB ROM, I want to try immediately leaving all compression off (LZ and Huffman) to see how much space that takes. Since ideally we'd compress as little as possible, it's faster to directly read uncompressed data.

PR #1345, opened 2026-01-29
Revise the final battles

Rangi42:


f-raZ0R:
Why a Machamp for Red?

PR #1356, opened 2026-02-06
Harsh sun

vulcandth:
# Harsh sun

This draft implements harsh sun as an overworld weather type with a slow, continuous fade between the normal palette state and a warm saturated state. The same weather also starts battles with sunlight.

The branch is rebased directly onto current `master`. The palette-swap work from #1608 is no longer duplicated here.

## Current behavior

- Harsh sun has no weather particles; its visual effect is palette-driven.
- BG and OBJ palettes use the normal overworld fade system, including fade catch-up for palettes loaded during an active transition.
- Text palettes and UI palettes that explicitly request daytime colors are not saturated.
- Copy-BG OBJ palettes reuse their already-adjusted BG source instead of applying saturation twice.
- Harsh sun pauses its pulse while scripts, Fly, map connections, or another overworld fade are active.
- Entering a battle during harsh sun starts `WEATHER_SUN` with the standard bright-sun message.
- The current temporary test override forces harsh sun on every outdoor `TOWN`, `ROUTE`, and `ISOLATED` map.

## Natural occurrence system already implemented

When the test override is disabled, harsh sun:

- is limited to daytime;
- unlocks after the Slowpoke Well plot;
- selects one Johto area and one Kanto area per day;
- treats connected or logically continuous map aliases as one area;
- persists its daily selections in save data.

The exact production activation policy is intentionally still pending approval.

## Performance and integration

- Uses the existing nonblocking overworld fade step instead of a separate `DelayFrames` loop.
- Performs cheap early returns when harsh sun is inactive, avoiding register-save and color-transform overhead during ordinary palette loads.
- Inlines the single-use per-color transform inside the shared BG/OBJ saturation routine.
- Reloads only active, already-assigned OBJ palette slots when the pulse reverses. It preserves mon-palette and object-glow metadata instead of clearing all dynamic palettes and scanning every object twice.
- Preserves master collision-driven object glows and applies harsh-sun state independently to both fade endpoints.
- Uses hardware color constants such as `PAL_COLORS`, `COLOR_CH_WIDTH`, `COLOR_CH_MAX`, `B_COLOR_GREEN`, and the channel masks instead of literal layout assumptions.
- Keeps persistent harsh-sun state inside an existing unused save-data gap so checksum-sensitive save addresses do not move.

## TODO

- [ ] Approve and implement the final production activation policy.
- [ ] Remove the `ALWAYS_HARSH_SUN_OUTDOORS` test override after the policy is approved.
- [x] Remove the known blocking and full-object-rescan sources of overworld slowdown.
- [ ] Confirm through extended gameplay testing that no perceptible slowdown remains on target hardware/emulators.
- [x] Deduplicate BG and OBJ saturation handling.
- [x] Make palette swaps and newly loaded OBJ palettes fade-aware.
- [x] Integrate with master object-glow palettes.
- [x] Rebase onto current master without carrying the squash-merged palette-swap commits.
- [x] Run `utils/optimize.py` and clean normal/debug builds.

## Suggested regression testing

- Walk and bike continuously in Goldenrod, Violet, and other object-heavy outdoor maps.
- Fly repeatedly between Goldenrod and Violet, checking arrival palettes and VRAM access.
- Cross normal/overcast and harsh-sun/non-harsh map connections in both directions.
- Open and close the Start menu, Town Map/Fly map, registered-item menu, and map-name sign during different pulse stages.
- Trigger lightning and time-of-day changes during or immediately after a pulse.
- Spawn or scroll an NPC into view mid-fade and verify that its palette appears at the current fade step.
- Check aquarium/campfire glow objects under harsh sun.
- Enter battles at both the normal and saturated ends of the pulse.


engine/gfx/color.asm
```diff
@@ -1030,7 +1030,135 @@ LoadMapPals:
...
+	add a
+	and $7c
+	ld d, a
+	ld a, c
+	srl a
+	srl a
+	srl a
+	and $03
+	or d
+	ld d, a
```
Rangi42: hardware.inc has many relevant constants for this! Including ones which you can `rept` by when e.g. doing N `add a` or `srl a` in a row.

engine/gfx/sprite_palettes.asm
```diff
@@ -322,6 +342,142 @@ CalculateStates:
...
+	or d
+	ld d, a
+
+	; write back (step hl back 2)
+	dec hl
+	ld [hl], d
+	dec hl
+	ld [hl], e
+	inc hl
+	inc hl
```
Rangi42: I'm not deeply inspecting this code, but it looks similar if not identical to the above code to combine R+G+B values into two GB bytes. And harsh sun isn't the only feature we have which does that. Might be a good opportunity to factor out a function for that.

engine/gfx/color.asm
vulcandth: Addressed on the rebased branch. The saturation code now uses the hardware color definitions (`PAL_COLORS`, `COLOR_CH_WIDTH`, `COLOR_CH_MAX`, `B_COLOR_GREEN`, `B_COLOR_BLUE`, and the channel masks), with `rept` counts derived from those constants instead of literal shift counts.

engine/gfx/sprite_palettes.asm
vulcandth: Addressed on the rebased branch. BG and OBJ callers now share one `ApplyHarshSunSaturationToPalette` implementation, so there is only one RGB unpack/adjust/repack path for harsh sun. The per-color body is inlined into that routine because it has no other caller, avoiding a `call`/`ret` pair for every transformed color.

Issue #871, opened 2023-08-09
Ruins of Alph's cave is different if seen from Violet


jamescastells:
This hack gladly fixed Ruins of Alph's map size to match with the rest of Johto, however, one problem remains:

If you go to the ruins and see the cave, it now has a different roof:

<img width="636" alt="image" src="https://github.com/Rangi42/polishedcrystal/assets/5696029/cf8d381e-85df-4456-a5d9-949291582118">

However, when seen from Violet City, the cave that's supposed to be the same (so the map fits), is seen with its regular roof:

<img width="635" alt="image" src="https://github.com/Rangi42/polishedcrystal/assets/5696029/39d8fabf-06a8-4404-9dd3-8b1e5f68aad6">

vulcandth:
Not enough blocks available to fix...

Rangi42:
Fixed by #1364.

PR #1364, opened 2026-02-10
Redesign Ruins of Alph more like HGSS

Rangi42:
image

### DIFF
--- a/maps/RuinsOfAlphOutside.asm
+++ b/maps/RuinsOfAlphOutside.asm
@@ -62,6 +62,11 @@ RuinsOfAlphOutside_MapScriptHeader:
 RuinsOfAlphOutsideOpenSinjohChamberScene:
 	checkevent EVENT_DO_RUINS_OF_ALPH_CLIMAX
 	iffalsefwd .End
+	sdefer RuinsOfAlphOutsideOpenSinjohChamberScript
+.End
+	end
+
+RuinsOfAlphOutsideOpenSinjohChamberScript:
 	showtext RuinsofAlphScientistClimax1Text
 	follow RUINSOFALPHOUTSIDE_SCIENTIST2, PLAYER
 	applymovement RUINSOFALPHOUTSIDE_SCIENTIST2, RuinsofAlphScientistClimaxApproachMovementData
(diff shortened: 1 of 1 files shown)


vulcandth:
This looks amazing! Can't wait to check it out in game.
Also, other than the Hyper Potion, do we want to add any other item balls for any items we have in game but haven't made available yet?

Rangi42:
IMO it's already a large quantity of items available on this map, including the Rock Smash ones.

Rangi42:
Bug: "That sounded like it came from over here..." prints twice during the cutscene where the Sinjoh chamber is opened.

Rangi42:
Fixes #871

Rangi42:
> Bug: "That sounded like it came from over here..." prints twice during the cutscene where the Sinjoh chamber is opened.

Possible diagnosis:

> **Root Cause**
> The bug is a stale joypad state interaction between `RunSceneScript` and `OWPlayerInput`.
>
> Here's the sequence on the first frame after warping to Ruins of Alph Outside:
>
> 1. `RunSceneScript` runs the scene script inline via `ScriptEvents`. The `showtext` command displays "That sounded like it came from over here…" and waits for an A press via `JoyWaitAorB`.
> 2. The player presses A to dismiss the text. `hJoyPressed` now has the A bit set.
> 3. The script continues to `follow` → `applymovement`, which calls `StopScript`, exiting `ScriptEvents`. `RunSceneScript` returns no carry (no `sdefer`).
> 4. `PlayerEvents` continues to `OWPlayerInput`, which calls `CheckAPressOW`. Since `hJoyPressed` still has A set from step 2, it proceeds to `TryObjectEvent`.
> 5. In the PR, **scientist2 is at position (19, 19)** — directly below the player at (19, 18) who is facing DOWN. `CheckFacingObject` finds the scientist!
> 6. `TryObjectEvent` calls `CallScript` with the scientist's script (`ObjectEvent` = just `end`). This overwrites `hScriptPos`, destroying the scene script's mid-execution position.
> 7. On the next frame, `RunSceneScript` fires again. Since `EVENT_DO_RUINS_OF_ALPH_CLIMAX` wasn't cleared yet (the scene script was interrupted before reaching `clearevent`), the scene script runs from the top again — showing the text a second time.
>
> **Why it worked on master**:
> On master, scientist2 was at (18, 18) — to the LEFT, not directly below the player. `CheckFacingObject` wouldn't find it facing DOWN.
>
> **Suggested Fix**
> Move scientist2 from (19, 19) to (20, 19) — one tile east, so it's no longer directly in front of the player's warp-in facing direction. Adjusted RuinsofAlphScientistClimaxApproachMovementData to add one extra step_left to compensate for the new starting position.

Rangi42:
Using `sdefer` fixed the bug. I believe this is the intended/correct method, so failing to defer a scene script doesn't count as a bug with the overworld engine.

PR #1388, opened 2026-02-13
Trick/Switcheroo fix.

aubsgp:
Previously, the check for if both items were empty was performed after the swapping to avoid clobbering a; the issue is that the swap itself also involves clobbering a. This caused Switcheroo to partially fail whenever the opponent lacked an item, even if the user happened to hold one, since you were effectively or-ing b with itself.

Since both parts clobber a anyway, I felt I might as well swap the order, in addition to fixing the bug.

### DIFF
--- a/engine/battle/move_effects/trick.asm
+++ b/engine/battle/move_effects/trick.asm
@@ -39,10 +39,6 @@ BattleCommand_trick:
 	ld a, b
 	ld [de], a
 
-	ld [hl], c
-	ld a, b
-	ld [de], a
-
 	ld hl, SwappedItemsText
 	call StdBattleTextbox
 


emilyploszaj (approved):

FredrIQ:
While this PR looks fine, wouldn't this also fix the issue (and save a byte)? Leaving the original code, but adding the 2nd line as follows:

	ld [hl], a
	or b ; z if none of the participants hold items
	ld a, b
	ld [de], a

Can't easily check at the moment, but if nobody replies here within a day, I'll add this fix later. Thanks for the PR!

aubsgp:
Oh, that's clever. Yeah, that should work too!

PR #1390, opened 2026-02-14
Scrolling initial options, and descriptions for in-game options

### LINKED ISSUE #1339: Scrollable Options Menu should allow "start" to exit
In addition to using "B" to save/exit the options menu, we should allow "start" as well.

### LINKED ISSUE #1386: Scrolling initial options, and descriptions for in-game options
The initial options have `(SEL>HELP)` descriptions, but are still paginated.

The in-game options scroll, but do not have descriptions.

Scrolling menus with helpful descriptions would be best for both.

vulcandth:
Resolves #1386
Resolves #1339

### DIFF
--- a/engine/menus/start_menu.asm
+++ b/engine/menus/start_menu.asm
@@ -43,7 +43,7 @@ StartMenu::
 
 .Reopen:
 	call RestoreSprites
-	call LoadWeatherPal
+	farcall LoadWeatherPal
 	call ClearSpritesUnderStartMenu
 	call UpdateSprites
 	call UpdateTimePals
@@ -129,9 +129,8 @@ StartMenu::
 	jr .ReturnEnd2
 
 .ReturnRedraw:
-	assert BANK("Dynamic Pals System") == BANK(@)
-	call ClearSavedObjPals
-	call DisableDynPalUpdates
+	farcall ClearSavedObjPals
+	farcall DisableDynPalUpdates
 	call ClearBGPalettes
 	call ExitMenu
 	call ReloadTilesetAndPalettes
--- a/layout.link
+++ b/layout.link
@@ -46,7 +46,6 @@ ROMX $03
 
 ROMX $04
 	"bank4"
-	"Dynamic Pals System" ; can float
 
 ROMX $05
 	"bank5"


Rangi42:
- [x] :bug: Bug: pressing Start to confirm your option choices plays `SFX_TRANSACTION` correctly, but pressing B or picking the Done option does not.

Rangi42:
- [x] :bug: Bug: when you press A to page through the description, the menu cursor briefly turns white while the description text gets updated. (If that's unavoidable scrolling menu functionality, we could cheat and overwrite the white arrow tile with the plain one, but that would be hacky.)

Rangi42:
- [x] :bug: Bug: `wOptionsMenuLastSelection` is not used properly, unclear if it's actually necessary.

Rangi42:
- [x] :bug: Bug: despite removing `wOptionsMenuValue` and `wOptionsMenuScrollPosition`, the Options menu still remembers how far down it had scrolled when you leave and re-enter it.

Rangi42:
- [x] :bug: Bug: entering and leaving the Options menu in-game corrupts the Start menu state, showing glitching Bug-Catching Contest UI. (I confirmed this happens even before my "`Some refactoring`" commit.)

Rangi42:
🔧 This needs updating now that #1359 has been merged.

vulcandth:
> * [ ]  🐛 Bug: `wOptionsMenuLastSelection` is not used properly, unclear if it's actually necessary.

`wOptionsMenuLastSelection` — IS necessary. It tracks the last-selected option so that `OptionsShared_ResetSelectionDescription` (function 3 callback, called every `ScrollingMenu` loop iteration) can skip resetting the description when the cursor hasn't moved. Without it, pressing A to page through a multi-page description would reset it to page 1 every time.

main.asm
```diff
@@ -529,11 +528,17 @@ INCLUDE "engine/gfx/fade.asm"
 
 SECTION "bank39", ROMX
 
-INCLUDE "engine/menus/options_menu.asm"
 INCLUDE "engine/movie/splash.asm"
 INCLUDE "engine/movie/intro.asm"
 
 
+SECTION "Options Menus", ROMX, BANK[$7d]
```
Rangi42: Why was this bank-specified?
```suggestion
SECTION "Options Menus", ROMX
```

home/text.asm
```diff
@@ -295,7 +298,18 @@ ContText::
 	pop de
 	jmp NextChar
 
+.stopHere
+	; Non-blocking: stop at this continuation point
+	ld a, 1
```
Rangi42: Always use `TRUE/FALSE` when appropriate (or at least `xor a ; FALSE`).
```suggestion
	ld a, TRUE
```
(We're probably at the point where PC should have an AGENTS.md file with that guideline. For another PR though!)

home/text.asm
```diff
@@ -325,7 +339,18 @@ Paragraph::
 	pop de
 	jmp NextChar
 
+.stopHere
+	; Non-blocking: stop at this paragraph break
+	ld a, 1
+	ldh [hStopPrintingString], a
+	pop hl
+	dec de
+	ret
```
Rangi42: Don't write the exact same thing twice -- that's especially important in home/ROM0. Just jump to the other one. (And of course then the other one should have an appropriate non-local label.)
```suggestion
```

home/text.asm
```diff
@@ -298,9 +298,8 @@ ContText::
 	pop de
 	jmp NextChar
 
-.stopHere
-	; Non-blocking: stop at this continuation point
-	ld a, 1
+Text_StopAtNonblockingPause::
```
Rangi42: This is assembly not text; please rename to just `StopAtNonblockingTextPause`.

Rangi42 (changes requested):

Rangi42:
@Copilot
```
error: The linker script assigns section "Dynamic Pals System" to address $7613, but then it would overflow ROMX by 192 bytes
    at layout.link(49)
```

PR #1408, opened 2026-02-25
Level Ranges for Wild Encounters

### LINKED ISSUE #1382: Refactor and generalize wild Pokémon data
Multiple goals here: increase capabilities of the wild encounter system, but also make it simpler/more compact/more failsafe to edit, and even save ROM space.

- [ ] Allow level ranges for each entry, instead of multiple same-species different-level entries.
- [ ] Allow multiple times of day for each entry (as a bitmask), instead of N entries for each time of day.
- [ ] Make Headbutt mons use the same kind of system, instead of the obfuscated [trainer ID-based system](https://polisheddex.vercel.app/tools/headbutt) that has individual "rare trees" and involves modular math. (See also #1463)

n0zep455:
Resolves #1382 

Removed CheckOnWater call from ChooseWildEncounter. This allows walking ("grass") encounters to be up to +4 levels above what's defined in def_grass_mons for the current map. This function was previously only available to surf encounters.

NOTE: The first commit in this branch on its own breaks game balance by allowing the possibility for over-leveled encounters and should not be merged before more work is performed. Level curve in data/wild/*_grass.asm files will require further update.

Future Plans: Update all wild encounters to function like the bug catching contest.
* We should already have the correct table from `data/wild`
* `CheckOnWater` call necessary to differentiate from "grass" tables
* Differentiating Day/Night relies on constant table lengths for the probability table.
* Can use use different columns for morn/day/nite and different level ranges for each time of day
* Examples:
```
def_grass_wildmons SPROUT_TOWER_2F
db 2 percent, 2 percent, 2 percent ; encounter rates: morn/day/nite
; morn%, min, max, day%, min, max, nite%, min, max, species
db  100,   3,   6,  100,   3,   6,     0,   3,   6, RATTATA ; first mon with 0 prob use  0
db   -1,   4,   6,   -1,   4,   6,   100,   4,   6, GASTLY  ; last  mon with 0 prob use -1
end_grass_wildmons

def_grass_wildmons SPROUT_TOWER_3F
db 2 percent, 2 percent, 2 percent ; encounter rates: morn/day/nite
; morn%, min, max, day%, min, max, nite%, min, max, species
db   40,   3,   5,   70,   3,   5,     0,   3,   5, RATTATA    ; Support multiple
db   40,   4,   6,   30,   3,   6,    40,   4,   6, RATTATA    ; mons defined for
db   20,   4,   6,   -1,   4,   6,    10,   4,   6, BELLSPROUT ; varied ranges
db   -1,   4,   6,   -1,   4,   6,    50,   4,   6, GASTLY     ; 
end_grass_wildmons
```
`wEncounterTimeID = wTimeOfDay * 3` (0->0, 1->3, 2->6)
`wEncounterLvLMin = wEncounterTimeID + 1`
`wEncounterLvLMax = wEncounterTimeID + 2`
`wEncounterPokMon = 9`

1. Load relevant wildmons table instead of probability table.
2. Random to traverse the table in column 0, 3, or 6
3. Random to pick number between min and max levels (1/2, 4/6, 7/8)
4. Load species in columm 9 with chosen level

### DIFF
--- a/data/wild/johto_grass.asm
+++ b/data/wild/johto_grass.asm
@@ -1,32 +1,12 @@
 ; Johto Pokémon in grass
 
-	def_grass_wildmons SPROUT_TOWER_2F
+	def_wildmons SPROUT_TOWER_2F
 	db 2 percent ; encounter rate
-	; morn
-	wildmon 3, RATTATA
-	wildmon 4, RATTATA
-	wildmon 5, RATTATA
-	wildmon 3, BELLSPROUT
-	wildmon 6, RATTATA
-	wildmon 5, BELLSPROUT
-	wildmon 6, BELLSPROUT
-	; day
-	wildmon 3, RATTATA
-	wildmon 4, RATTATA
-	wildmon 5, RATTATA
-	wildmon 3, BELLSPROUT
-	wildmon 6, RATTATA
-	wildmon 5, BELLSPROUT
-	wildmon 6, BELLSPROUT
-	; nite
-	wildmon 3, RATTATA
-	wildmon 4, RATTATA
-	wildmon 5, GASTLY
-	wildmon 3, BELLSPROUT
-	wildmon 6, GASTLY
-	wildmon 5, BELLSPROUT
-	wildmon 6, BELLSPROUT
-	end_grass_wildmons
+	;       morn%, min, max, day%, min, max, nite%, min, max, species
+	wildmon    85,   3,   6,   85,   3,   6,    65,   3,   4, RATTATA
+	wildmon    15,   3,   6,   15,   3,   6,    15,   3,   6, BELLSPROUT
+	wildmon    -1,   3,   6,   -1,   3,   6,    20,   4,   6, GASTLY
+	end_wildmons
 
 	def_grass_wildmons SPROUT_TOWER_3F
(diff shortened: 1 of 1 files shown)


vulcandth:
I believe this PR misses the intent of the issue:

> Refactoring from fixed-size lists to flag/range-based data could save space and be easier to read/edit.

The intent is to save space, not just to vary the levels using the existing system.

n0zep455:
Understood.

I also left a comment on the issue about refactoring all wild encounters to work like the bug catching contest. If that sounds good I can start working towards that.

vulcandth:
I'll let @Rangi42 weigh in on this. Do really appreciate the contribution, but I don't want you to put in a bunch of work if its not the direction we want to go in. :)

n0zep455:
Looked a little more thoroughly at the polished implementation for wildmon/contest_mon, added some to the initial PR comment and making some broken changes for demonstrative purposes.

Rangi42:
Thanks for this PR, but I think part of the issue is still "ironing out exactly what sort of capabilities we want in the data structure", so there isn't a specific task spec ready yet to be implemented. (For example, can/should times of day share mutual entries, should evening have its own entries, etc.)

PR #1411, opened 2026-02-28
Refactor VBlank0 routine calls

vulcandth:
Problem:

vBlank was skipping many routines that should be able to safely run in the time given. This was causing issues where if i'm fading palettes in the overworld, the tileset animations for example would run super slow. This allows more to run.. since we are in double speed we have the time.

I did some testing in-game with inaccessible vram enabled in BGB and only had 1 hit, which was unrelated due to when `PlaceMapNameSign:` runs and calls `HDMATransfer_OnlyTopFourRows:` which disables interrupts. When it finally re-enables interrupts we jump to `vBlank::` but the timeframe for vblank has since passed.

### DIFF
--- a/home/vblank.asm
+++ b/home/vblank.asm
@@ -162,9 +162,7 @@ VBlank0::
 	; At cp 149: skips when ≥1140 M-cycles elapsed (buffer updates).
 	ldh a, [rLY]
 	cp 149
-	jr nc, .skipBGMap
-	call UpdateBGMap
-.skipBGMap:
+	call c, UpdateBGMap
 
 	; Tile data transfers (have their own LY checks).
 	call Serve2bppRequest


Rangi42:
> I did some testing in-game with inaccessible vram enabled in BGB and only had 1 hit, which was unrelated due to when `PlaceMapNameSign:` runs and calls `HDMATransfer_OnlyTopFourRows:` which disables interrupts. When it finally re-enables interrupts we jump to `vBlank::` but the timeframe for vblank has since passed.

Does this mean that `PlaceMapNameSign` is capable of accessing VRAM outside VBlank even in the current `master`? Is that reproducible in some way? Because I have "Inaccessible VRAM: Break on access" enabled in BGB, but haven't seen it happen yet.

(Edit: Although I also haven't managed to trigger it in this PR build either.)

home/vblank.asm
```diff
@@ -140,27 +140,38 @@ VBlank0::
...
+	call PlaceFootprints ; 11 [empty] / 53 [2 footprints] M-cycles worst case
+
+	; Heavy BG map copy (~909 M-cycles). Gate on LY to avoid
+	; overrunning VBlank when earlier operations consumed time.
+	; At cp 149: skips when ≥1140 M-cycles elapsed (buffer updates).
+	ldh a, [rLY]
+	cp 149
+	jr nc, .skipBGMap
+	call UpdateBGMap
+.skipBGMap:
```
Rangi42:
```suggestion
	call c, UpdateBGMap
```

home/vblank.asm
```diff
@@ -140,27 +140,38 @@ VBlank0::
...
 	call Serve2bppRequest
 	call Serve1bppRequest
-	call AnimateTileset
-	call PlaceFootprints
 
-.done
-	call UpdateCGBPalsLYTimed
-	call PushOAM
+	; OAM DMA — must complete within VBlank.
+	call PushOAM ; 9 [skip] / 185 [fire]
```
Rangi42, line 172: Are these "skip / fire" comments understood and accurate? Because I don't see how `PushOAM` could possibly *be* "skip"ped in this control flow. Are they just leftover notetaking by Codex?

vulcandth:
> > I did some testing in-game with inaccessible vram enabled in BGB and only had 1 hit, which was unrelated due to when `PlaceMapNameSign:` runs and calls `HDMATransfer_OnlyTopFourRows:` which disables interrupts. When it finally re-enables interrupts we jump to `vBlank::` but the timeframe for vblank has since passed.
> 
> Does this mean that `PlaceMapNameSign` is capable of accessing VRAM outside VBlank even in the current `master`? Is that reproducible in some way? Because I have "Inaccessible VRAM: Break on access" enabled in BGB, but haven't seen it happen yet.
> 
> (Edit: Although I also haven't managed to trigger it in this PR build either.)

PlaceMapNameSign runs fine.. but it has a chance to delay vblank from running during vblank... causeing it's routines (vblank) to ~~break~~ try to write outside vblank. Doesn't happen all the time.

home/vblank.asm
vulcandth: No I wrote those. The skip is talking about the "work needed?" guard.

Although I think i'm off a cycle...  `ldh a, [c]` is 2 cycles not 3.

```
PushOAM::
	ldh a, [hOAMUpdate]
	and a
	ret nz
```

vulcandth: The comments are not completely necessary.... There could be miscounting on the cycles.. and obviously if those routines get refactored the comments will become outdated... but they do give a general idea of how long the routine could take.

PR #1508, opened 2026-03-21
Remove black corners from items

### LINKED ISSUE #1491: Add black corner pixels to item icons when necessary, instead of removing them
Every item icon has its four corners colored with black pixels, so that the Bag UI's border has rounded corners. However, those black pixels need to get removed for display in overworld textboxes. (See `WhiteOutDecompressedItemIconCorners`.)

It might be better to store images with white corners, and add black pixels to them when inside the Bag.

- Advantage: White corners should LZP compress better, taking up less ROM. It's also easier to avoid forgetting to color in the corners, which has happened with some icons.
- Disadvantage: It might slow down scrolling through the Bag pockets. If this is significant, we should leave the icons as-is.

leo-petrucci:
Closes #1491 

This PR does a couple of things:
- Removes all pre-existing black pixel in the corners of the items thumbnails (this was done with an automated script 👀)
- Implements a function to manually re-add black pixels when the item preview is loaded in the bag
- Simplifies the functions for both cases that had to manually _white out_ the black pixels to just load the item thumbnails

Changes were smoke tested, previews work in both dialog boxes and in the summary screens.

Screenshot 2026-03-21 at 20 53 40

Screenshot 2026-03-21 at 20 54 29

There's no visible slowdown when scrolling the bag.

https://github.com/user-attachments/assets/c105b134-025d-494b-82b9-fb24c87710ec


engine/gfx/item_icons.asm
```diff
@@ -108,37 +105,37 @@ _SetupLoadItemOrKeyItemIcon:
...
 	ret
 
-WhiteOutDecompressedItemIconCorners:
+BlackOutDecompressedItemIconCorners:
 	call RunFunctionInWRA6
 .Function:
-	lb bc, %01111111, %11111110
-	ld hl, wDecompressScratch tile 0
+	lb bc, %10000000, %00000001 
+	ld hl, wDecompressScratch tile 0
```
Rangi42:
```suggestion
	ld hl, wDecompressScratch tile 0
```

engine/gfx/item_icons.asm
```diff
@@ -108,37 +105,37 @@ _SetupLoadItemOrKeyItemIcon:
...
 	ld a, [hl]
-	and b
+	or b 
 	ld [hli], a
 	ld a, [hl]
-	and b
+	or b 
 	ld [hl], a
-	ld hl, wDecompressScratch tile 2
+		ld hl, wDecompressScratch tile 2
```
Rangi42:
```suggestion
	ld hl, wDecompressScratch tile 2
```

Rangi42:
Nice use of automation here, thank you! The tested results look good. :)

Rangi42:
This also saved 1270 bytes of ROM!

PR #1518, opened 2026-03-30
Remove the "predef" system, replacing `predef/predef_jump` with `farcall/farjp`

Rangi42:
Farcalls have long since preserved the same registers as predefs, so we don't really need the predef system any more. The question is whether it's worth keeping for shorter calls.

A `predef` or `predef_jump` used 2 bytes. A `farcall` or `farjp` uses 4 bytes. We had 238 of them (via `git grep -nP '^\tpredef(_jump)? ' | wc -l`), so this would have used an extra 476 bytes, but then freed bytes by removing the predef `rst` and handler in ROM0 and the pointer table in ROMX.

Total free space for `make debug` went from 31477 to 31165 bytes, so we **spent 312 more bytes** overall. However, free space in ROM0 went from 170 to 216 bytes, so we **saved 46 home bytes**, and **freed up `rst $38`** for whatever we want. (Note that `rst $38` can be ["called conditionally"](https://github.com/pret/pokecrystal/wiki/Optimizing-assembly-code#conditional-rst-38)!)

There are 28 "unnecessary `farcall`" warnings, which I didn't bother dealing with to keep the diff to the essentials. That's also why I didn't rename `FlagPredef` or `FrontpicPredef`.


Rangi42:
I left `predef/predef_jump` awareness in tools like optimize.py so they'll continue to be useful for other branches and repos.

Rangi42:
These are our top 5 `farcall`/`farjp`-ed functions:
```
     75 Sine
     36 Cosine
     30 Multiply
     30 Divide
     19 EngineFlagAction
```
So we might be best off not even using `rst $38` and just doing `di :: ld a, ERR_RST_38 :: jmp Crash`, like with `rst $00`.

PR #1544, opened 2026-04-20
Teleport guy always there

RyanRutkin:
The Teleport Guy inside of the Indigo Plateau PokeCenter will no longer disappear after encountering Lawrence in Vermillion City.


RyanRutkin:
Wrong base

PR #1566, opened 2026-05-26
Add Bloodmoon Ursaluna party configurations for battle facilities

NickF56:


FredrIQ:
Looks fine, but a save patch is required after adding this, to fix the Battle facility save data (particularly for the Factory).

NickF56:
Bloodmoon Ursa can probably wait to join the ranks of the battle tower teams until Vulc has time to prep a save patch, then. 

@FredrIQ if you have time to make it so battle tower sets can have optimised 510 EV spreads, I’d be happy to add those EV spreads

PR #1572, opened 2026-06-09
Optional level caps

albertquiroga:
## Add optional level cap

Adds a new toggle to the initial options menu — **Level cap: Yes / No** (Default: No)— that enforces an experience cap. When enabled, Pokémon stop gaining Exp once they reach the level of the next Gym Leader's or Champion's ace. The cap rises automatically as badges are earned.

### How the cap is determined

The cap is computed by `GetLevelCap` (added to `engine/overworld/wildmons.asm`) using the following priority:

| Condition | Cap |
|---|---|
| Beaten Red | 100 |
| Beaten Kanto Elite Four | 90 |
| Otherwise | `LevelCaps[badge_count]` |

The `LevelCaps` table (added as `data/wild/level_caps.asm`) maps badge count to the next major boss's ace level:

| Badges | Cap | Next boss |
|---|---|---|
| 0 | 13 | Falkner |
| 1 | 17 | Bugsy |
| 2 | 21 | Whitney |
| 3 | 26 | Morty |
| 4 | 31 | Chuck |
| 5 | 37 | Jasmine |
| 6 | 42 | Pryce |
| 7 | 47 | Clair / Johto E4 |
| 8 | 60 | Lt. Surge |
| 9 | 62 | Sabrina |
| 10 | 64 | Misty |
| 11 | 65 | Erika |
| 12 | 65 | Janine |
| 13 | 66 | Brock |
| 14 | 69 | Blaine |
| 15 | 70 | Blue / Kanto E4 |
| 16 | 80 | (post-game) |

This logic has been based on the nuzlock-style caps defined [here](https://github.com/Rangi42/polishedcrystal/wiki/Nuzlocke#list-of-highest-levels)

### Files changed

- **`constants/ram_constants.asm`** — adds `LEVEL_CAP_OPT` constant (bit 5 of `wInitialOptions2`)
- **`ram/wram0.asm`** — adds `wLevelCap` scratch variable in WRAM0; documents `LEVEL_CAP_OPT` in the `wInitialOptions2` bit layout
- **`data/wild/level_caps.asm`** *(new)* — the badge-indexed level cap table
- **`engine/overworld/wildmons.asm`** — adds `GetLevelCap` function; includes new table
- **`engine/battle/core.asm`** — `GiveExperiencePoints` calls `GetLevelCap` once per battle and skips Exp for any Pokémon at or above the cap
- **`engine/items/item_effects.asm`** — Rare Candy and Exp. Candy both respect the cap; at or above the cap, Rare Candy still triggers level-up evolutions (matching vanilla level 100 behaviour) but the candy is not consumed
- **`engine/menus/initial_options_menu.asm`** — adds `InitialOptions_LevelCap` routine and wires it into the option dispatch table; bumps `NUM_INITIAL_MENU_OPTIONS` to 12
- **`data/options/initial_option_names.asm`** — adds "Level cap" entry
- **`data/options/initial_options_descriptions.asm`** — adds full in-game description for the option

### Design notes

- When the option is **off**, `GetLevelCap` returns `MAX_LEVEL` (100), preserving vanilla behaviour exactly — no other code path is affected.
- The cap is cached in `wLevelCap` at the start of each Exp distribution loop and each candy use, avoiding repeated `farcall`s per party member.
- The cap uses `>= cap` (not `== 100`) so Pokémon already over-levelled from a previous session are also blocked from gaining further Exp.

### Considerations

I am aware of the [history with nuzlock modes](https://github.com/Rangi42/polishedcrystal/blob/master/FAQ.md#what-happened-to-nuzlocke-mode) this romhack has and therefore the potential pushback this PR might have, however I believe level caps are a feature more related to enhancing the difficulty, level curve and overall experience of the romhack than they are to nuzlocking. Plus, this is something the game can easily handle that removes a lot of headaches and manual work from the player, cause even if the player wants to self-impose level caps they have to be constantly juggling pokemon to play the "do not exceed the level" roulette game rather than enjoying the rom for what it is. The option is also disabled by default and the player can choose to enable it when starting a new game.

I will also add that the changes in this PR have been mostly written by Claude and steered by a human - I want to make that very clear as I am aware a lot of GH repo owners do not appreciate AI-written contributions, not sure if that is the case here.

### DIFF
--- a/constants/ram_constants.asm
+++ b/constants/ram_constants.asm
@@ -118,7 +118,8 @@ DEF LINK_OPTMASK EQU (1 << NATURES_OPT) | (1 << ABILITIES_OPT) | (1 << PERFECT_I
 	const NO_EXP_OPT           ; 2
 	const RTC_OPT              ; 3
 	const EVOLVE_IN_BATTLE_OPT ; 4
-	const_skip 2
+	const LEVEL_CAP_OPT        ; 5
+	const_skip 1
 	const RESET_INIT_OPTS      ; 7
 
 	const_def
--- a/data/options/initial_option_names.asm
+++ b/data/options/initial_option_names.asm
@@ -11,6 +11,7 @@ InitialOptionNames:
 	dw .TradedMon
 	dw .EvolveInBattle
 	dw .ColorVariation
+	dw .LevelCap
 	assert_table_length NUM_INITIAL_MENU_OPTIONS
 
 .Natures:
@@ -35,3 +36,5 @@ InitialOptionNames:
 	db "Evolve in battle@"
 .ColorVariation:
 	db "Color variation@"
+.LevelCap:
+	db "Level cap@"
--- a/data/options/initial_options_descriptions.asm
+++ b/data/options/initial_options_descriptions.asm
@@ -13,6 +13,7 @@ InitialOptionDescriptions:
 	dw .TradedMon
 	dw .EvolveInBattle
 	dw .ColorVariation
+	dw .LevelCap
 	assert_table_length NUM_INITIAL_MENU_OPTIONS
 
 .Natures:
@@ -162,6 +163,22 @@ InitialOptionDescriptions:
 
 	para "variation based"
 	line "on nicknames."
+	prompt
+
+.LevelCap:
+	text "Your #mon stop"
+	line "gaining Exp. once"
+	cont "they reach the"
+
+	para "level of the next"
+	line "Gym Leader's or"
+	cont "Champion's ace."
+
+	para "The cap rises as"
+	line "you earn Badges."
+
+	para "Exp.Candy and Rare"
+	line "Candy obey it too."
 	prompt
 
 InitialOptionsDoneDescription:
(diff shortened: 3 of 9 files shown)


Rangi42:
Firstly, thank you for submitting code here; it goes above and beyond a feature request. AI-assisted code is fine as long as the person behind it is still taking responsibility for its output -- putting in the same amount of comprehension and debugging and quality assurance as they would if they'd written it from scratch.

(The code here looks "fine". I'd have some change requests to it, but it looks comparable to a hand-written feature PR. Also, in my experience, having *something* as a basis makes it easier to reach a good endpoint, even if the original ends up getting refactored/rewritten.)

However, level caps aren't going to be officially supported as an option. I think the particular level values in your table make it clear why: they were never chosen with the expectation that players would or should stay at/below each "boss" trainer's level. Like, is it really practical to not gain any levels between Erika and Janine? There's also considerable subjectivity in which trainers count as "bosses": you could argue that one or both rivals or the Team Rocket admins should count here.

PR #1580, opened 2026-06-30
Use `text_farend` instead of `text_far`+`text_end`

Rangi42:
This uses a high bit flag like `farjp`, saving 655 bytes.


Rangi42:
For even more space, many of these could just put the text in the actual code file and avoid any far bank.

PR #1608, opened 2026-08-28
Make palette swaps fade-aware

vulcandth:
## Summary

Make map `paletteswap` entries integrate with the overworld fade system instead of forcing an immediate full palette application.

This prevents swapped BG palettes from snapping during weather, time-of-day, map-connection, lightning, and similar fades. It also makes a palette that changes or first appears partway through a fade start from the correct previous-state colors and catch up to the current fade step immediately.

## Palette state handling

- Make `SwapColorPalette` use the fade system's selected `wPalState` rather than reading the live time-of-day and overcast values directly.
- Write previous-state palettes to `wBGPals2`, the active/fading buffer, and current-state palettes to `wBGPals1`, the destination buffer.
- Track each palette-swap rectangle's current inside/outside state and whether that entry has ever been initialized. `wPaletteSwapFlag` uses:
  - low nibble: current inside/outside state for entries 1-4;
  - high nibble: initialization state for entries 1-4.
- Add named constants for the temporary inside/changed processing flags instead of relying on numeric bit positions.
- Add a build-time assertion limiting a palette-swap list to four entries, matching the capacity that already exists in `wPaletteSwapFlag`.

## Fade-aware updates and catch-up

- Avoid the old forced full-palette update while an overworld fade is active.
- During an active fade, process palette swaps on the movement frame where the player's rectangle state can change, while skipping unnecessary work on idle fade frames.
- Catch up only destination BG palette slots affected by an initialized or changed entry.
- Rebuild a changed slot under the fade's previous conditions, load its destination under the current conditions, and replay the elapsed fade steps for that one BG palette.
- Support fades whose previous state is white by initializing the changed active BG palette to white before catch-up.

Palette-swap lists are resolved in two passes. The first pass updates all rectangle states and records which BG palette slots were affected. The second pass applies entries in list order. This is required for chained entries such as Goldenrod City's two `PAL_BG_WATER` swaps, where a later `NULL` palette means “leave the earlier entry for this slot intact.” The final effective palette now receives catch-up even when the entry whose rectangle changed selects `NULL`.

No additional WRAM or persistent palette-swap state is introduced for this resolution; the affected BG slots are held in a transient register mask.

## Shared fade helpers and copied OBJ palettes

- Generalize palette fade catch-up so it can target either an OBJ or BG palette.
- Set `hCGBPalUpdate` after catching up a BG palette so its active colors are uploaded normally.
- Make copy-BG OBJ palettes select the active or destination BG buffer according to `wPalState`.
- Do not replay fade steps for ordinary copy-BG OBJ palettes that already copied an in-progress active BG palette; this avoids advancing them ahead of the BG palette they mirror.
- Preserve normal catch-up for white-origin fades and `PAL_OW_COPY_BG_WHITE`, which are constructed from destination colors.
- Mask `wPaletteSwapFlag` to its current-state nibble when choosing the Fly/Cut leaf palette, so initialization bits are not mistaken for active swaps.

## VBlank safety

`DI_DelayFrame` can wait through VBlank with interrupts disabled during HBlank DMA. Clear the stale VBlank request before returning so it is not serviced later in the visible frame, where a deferred CGB palette upload could write inaccessible palette VRAM.

## Performance and size

- Skip catch-up unless a palette-swap entry actually changed.
- Skip the added fade-state query and destination-mask calculation on the common no-change path.
- Move the larger palette rebuild/catch-up helper into the banked color engine to keep the ROM0 `Home` section within capacity.
- Use a tail `farjp` for the final BG catch-up call.
- `python3 utils/optimize.py` reports zero findings across all assembly files changed by this branch.

## Validation

- `make clean && make -j6`
- `make clean && make -j6 debug`
- Intentional five-entry palette-swap list fails assembly with: `A palette swap list may have at most 4 entries`.
- Reviewed chained same-slot swaps, selected `NULL` entries, first-time initialization, ordinary and white-origin fades, copy-BG OBJ palettes, ROM/WRAM bank restoration, and stack balance on every exit path.

### DIFF
--- a/constants/gfx_constants.asm
+++ b/constants/gfx_constants.asm
@@ -63,6 +63,15 @@ DEF DEXTILE_FROM_DEXMAP EQU 1 << DEXTILE_FROM_DEXMAP_F
 DEF NUM_FLYFROM_ANIM_OAMS EQU 4 + 9 ; 4 for flymon, 9 for leaves
 DEF NUM_FLYTO_ANIM_OAMS   EQU 4 + 8 ; 4 for flymon, 8 for leaves
 
+; wPaletteSwapFlag uses one nibble each for current and initialized states.
+DEF NUM_PALETTE_SWAP_STATES EQU 4
+DEF PALETTE_SWAP_STATE_MASK EQU (1 << NUM_PALETTE_SWAP_STATES) - 1
+
+; Temporary flags used while processing one palette swap entry.
+DEF PALETTE_SWAP_INSIDE_F  EQU 0
+DEF PALETTE_SWAP_CHANGED_F EQU 7
+DEF PALETTE_SWAP_INSIDE    EQU 1 << PALETTE_SWAP_INSIDE_F
+
 ; wPalState types (see engine/gfx/sprite_palettes.asm)
 	const_def
 	const PREV_PALSTATE ; 0
--- a/constants/gfx_constants.asm
+++ b/constants/gfx_constants.asm
@@ -63,10 +63,6 @@ DEF DEXTILE_FROM_DEXMAP EQU 1 << DEXTILE_FROM_DEXMAP_F
 DEF NUM_FLYFROM_ANIM_OAMS EQU 4 + 9 ; 4 for flymon, 9 for leaves
 DEF NUM_FLYTO_ANIM_OAMS   EQU 4 + 8 ; 4 for flymon, 8 for leaves
 
-; wPaletteSwapFlag uses one nibble each for current and initialized states.
-DEF NUM_PALETTE_SWAP_STATES EQU 4
-DEF PALETTE_SWAP_STATE_MASK EQU (1 << NUM_PALETTE_SWAP_STATES) - 1
-
 ; Temporary flags used while processing one palette swap entry.
 DEF PALETTE_SWAP_INSIDE_F  EQU 0
 DEF PALETTE_SWAP_CHANGED_F EQU 7
--- a/engine/gfx/color.asm
+++ b/engine/gfx/color.asm
@@ -1059,20 +1059,25 @@ SwapColorPalette::
 	add hl, de
 
 	; Previous-state palettes become active; current-state palettes are the target.
-	push hl
-	ld hl, wBGPals2
+	; d = [wPalState] == PREV_PALSTATE ? LOW(wBGPals2) : LOW(wBGPals1)
 	ld a, [wPalState]
 	assert PREV_PALSTATE == 0
 	and a
+	ld d, LOW(wBGPals2)
 	jr z, .got_target
-	ld hl, wBGPals1
+	ld d, LOW(wBGPals1)
 .got_target
+	; Set `de` to the `b`th palette of `wBGPals1` or `wBGPals2`.
 	ld a, b
-	ld bc, 1 palettes
-	rst AddNTimes
-	ld d, h
-	ld e, l
-	pop hl
+	add a
+	add a
+	add a
+	add d ; LOW(wBGPals1) or LOW(wBGPals2)
+	ld e, a
+	assert HIGH(wBGPals1) == HIGH(wBGPals2)
+	adc HIGH(wBGPals1)
+	sub e
+	ld d, a
 
 	; Skip past the regular palettes to the overcast ones if applicable.
 	ld a, PALSTATE_OVERCAST_INDEX
@@ -1088,20 +1093,16 @@ SwapColorPalette::
 	jmp FarCopyColorWRAM
 
 UpdatePaletteSwapState::
-; wPaletteSwapFlag's low nibble stores the current state of each entry;
-; its high nibble records which entries have been initialized.
+; wPaletteSwapStates stores the current state of each entry.
+; wPaletteSwapInits records which entries have been initialized.
 ; Input: `c` PALETTE_SWAP_INSIDE_F set inside the rectangle,
 ;        `e` = this entry's state bit.
 ; Output: `c` PALETTE_SWAP_CHANGED_F set on first check or a state change.
-	ld a, [wPaletteSwapFlag]
-	ld b, a
-	; Mark this entry initialized, and build its new state in `a`.
-	ld a, e
-	swap a
-	or b
+	ld a, [wPaletteSwapStates]
+	ld d, a
+	; Build this entry's new state in `a`.
 	bit PALETTE_SWAP_INSIDE_F, c
 	jr nz, .inside
-	ld d, a
 	ld a, e
 	cpl
 	and d
@@ -1109,12 +1110,19 @@ UpdatePaletteSwapState::
 .inside
 	or e
 
-	; Store only if the initialized or current state changed.
+	; Store only on first check or if the current state changed.
 .compare
-	xor b
-	ret z
-	xor b
-	ld [wPaletteSwapFlag], a
+	cp d
+	jr nz, .changed
+	ld a, [wPaletteSwapInits]
+	and e
+	ret nz
+	ld a, d
+.changed
+	ld [wPaletteSwapStates], a
+	ld a, [wPaletteSwapInits]
+	or e
+	ld [wPaletteSwapInits], a
 	set PALETTE_SWAP_CHANGED_F, c
 	ret
 
--- a/engine/gfx/color.asm
+++ b/engine/gfx/color.asm
@@ -1117,9 +1117,10 @@ UpdatePaletteSwapState::
 	ld a, [wPaletteSwapInits]
 	and e
 	ret nz
-	ld a, d
+	jr .finish
 .changed
 	ld [wPaletteSwapStates], a
+.finish
 	ld a, [wPaletteSwapInits]
 	or e
 	ld [wPaletteSwapInits], a
@@ -1134,12 +1135,18 @@ CatchUpPaletteSwapFade::
 	ld a, [wPalWhiteState]
 	and a
 	jr z, .swap_previous
-	ld hl, wBGPals2
+
+	; de = wBGPals2 palette b
 	ld a, b
-	ld bc, 1 palettes
-	rst AddNTimes
-	ld d, h
-	ld e, l
+	add a
+	add a
+	add a
+	add LOW(wBGPals2)
+	ld e, a
+	adc HIGH(wBGPals2)
+	sub e
+	ld d, a
+
 	farcall CopyWhitePal
 	jr .got_previous
 
@@ -1147,14 +1154,14 @@ CatchUpPaletteSwapFade::
 	xor a
 	assert PREV_PALSTATE == 0
 	ld [wPalState], a
+	push bc
 	call SwapColorPalette
+	pop bc
 
 .got_previous
 	pop de
-	pop bc
 	ld a, CURR_PALSTATE
 	ld [wPalState], a
-	push bc
 	call SwapColorPalette
 	pop bc
 	ld a, b
--- a/engine/gfx/color.asm
+++ b/engine/gfx/color.asm
@@ -1086,6 +1086,46 @@ SwapColorPalette::
 	ld bc, 1 palettes
 	jmp FarCopyColorWRAM
 
+UpdatePaletteSwapState::
+; Input: `c` bit 0 set inside the rectangle, `e` = this entry's state bit.
+; Output: `c` bit 7 set on first check or a state change.
+	ld a, e
+	swap a
+	ld d, a
+	ld a, [wPaletteSwapFlag]
+	and d
+	jr z, .changed
+	ld a, [wPaletteSwapFlag]
+	and e
+	jr z, .was_outside
+	bit 0, c
+	ret nz
+	jr .changed
+.was_outside
+	bit 0, c
+	ret z
+
+.changed
+	set 7, c
+	ld a, [wPaletteSwapFlag]
+	or d
+	ld [wPaletteSwapFlag], a
+	bit 0, c
+	jr z, .clear_state
+	ld a, [wPaletteSwapFlag]
+	or e
+	ld [wPaletteSwapFlag], a
+	ret
+
+.clear_state
+	ld a, e
+	cpl
+	ld d, a
+	ld a, [wPaletteSwapFlag]
+	and d
+	ld [wPaletteSwapFlag], a
+	ret
+
 INCLUDE "data/tileset_palettes.asm"
 
 INCLUDE "data/maps/environment_colors.asm"
--- a/engine/gfx/color.asm
+++ b/engine/gfx/color.asm
@@ -1087,43 +1087,34 @@ SwapColorPalette::
 	jmp FarCopyColorWRAM
 
 UpdatePaletteSwapState::
-; Input: `c` bit 0 set inside the rectangle, `e` = this entry's state bit.
-; Output: `c` bit 7 set on first check or a state change.
+; wPaletteSwapFlag's low nibble stores the current state of each entry;
+; its high nibble records which entries have been initialized.
+; Input: `c` PALETTE_SWAP_INSIDE_F set inside the rectangle,
+;        `e` = this entry's state bit.
+; Output: `c` PALETTE_SWAP_CHANGED_F set on first check or a state change.
+	ld a, [wPaletteSwapFlag]
+	ld b, a
+	; Mark this entry initialized, and build its new state in `a`.
 	ld a, e
 	swap a
+	or b
+	bit PALETTE_SWAP_INSIDE_F, c
+	jr nz, .inside
 	ld d, a
-	ld a, [wPaletteSwapFlag]
-	and d
-	jr z, .changed
-	ld a, [wPaletteSwapFlag]
-	and e
-	jr z, .was_outside
-	bit 0, c
-	ret nz
-	jr .changed
-.was_outside
-	bit 0, c
-	ret z
-
-.changed
-	set 7, c
-	ld a, [wPaletteSwapFlag]
-	or d
-	ld [wPaletteSwapFlag], a
-	bit 0, c
-	jr z, .clear_state
-	ld a, [wPaletteSwapFlag]
-	or e
-	ld [wPaletteSwapFlag], a
-	ret
-
-.clear_state
 	ld a, e
 	cpl
-	ld d, a
-	ld a, [wPaletteSwapFlag]
 	and d
+	jr .compare
+.inside
+	or e
+
+	; Store only if the initialized or current state changed.
+.compare
+	xor b
+	ret z
+	xor b
 	ld [wPaletteSwapFlag], a
+	set PALETTE_SWAP_CHANGED_F, c
 	ret
 
 INCLUDE "data/tileset_palettes.asm"
--- a/engine/gfx/color.asm
+++ b/engine/gfx/color.asm
@@ -1062,7 +1062,8 @@ SwapColorPalette::
 	push hl
 	ld hl, wBGPals2
 	ld a, [wPalState]
-	and a ; PREV_PALSTATE == 0
+	assert PREV_PALSTATE == 0
+	and a
 	jr z, .got_target
 	ld hl, wBGPals1
 .got_target
--- a/engine/gfx/color.asm
+++ b/engine/gfx/color.asm
@@ -1047,8 +1047,9 @@ SwapColorPalette::
 	or e
 	ret z
 
-	; Set `hl` to the `[wTimeOfDayPal]`th palette in the table at `de`.
-	ld a, [wTimeOfDayPal]
+	; Set `hl` to the selected state's time-of-day palette in the table at `de`.
+	ld a, PALSTATE_TIME_OF_DAY
+	farcall GetPalState
 	and 3
 	add a
 	add a
@@ -1057,23 +1058,24 @@ SwapColorPalette::
 	ld h, 0
 	add hl, de
 
-	; Set `de` to the `b`th palette of `wBGPals1`.
+	; Previous-state palettes become active; current-state palettes are the target.
+	push hl
+	ld hl, wBGPals2
+	ld a, [wPalState]
+	and a ; PREV_PALSTATE == 0
+	jr z, .got_target
+	ld hl, wBGPals1
+.got_target
 	ld a, b
-	add a
-	add a
-	add a
-	add LOW(wBGPals1)
-	ld e, a
-	adc HIGH(wBGPals1)
-	sub e
-	ld d, a
+	ld bc, 1 palettes
+	rst AddNTimes
+	ld d, h
+	ld e, l
+	pop hl
 
 	; Skip past the regular palettes to the overcast ones if applicable.
-	push hl
-	push de
-	farcall GetOvercastIndex
-	pop de
-	pop hl
+	ld a, PALSTATE_OVERCAST_INDEX
+	farcall GetPalState
 	and a
 	jr z, .not_overcast
 	ld bc, 4 palettes
--- a/engine/gfx/color.asm
+++ b/engine/gfx/color.asm
@@ -1118,6 +1118,40 @@ UpdatePaletteSwapState::
 	set PALETTE_SWAP_CHANGED_F, c
 	ret
 
+CatchUpPaletteSwapFade::
+; Rebuild BG palette `b` from palette list `de` under the fade's previous and
+; current conditions, then catch its active palette up to the current step.
+	push bc
+	push de
+	ld a, [wPalWhiteState]
+	and a
+	jr z, .swap_previous
+	ld hl, wBGPals2
+	ld a, b
+	ld bc, 1 palettes
+	rst AddNTimes
+	ld d, h
+	ld e, l
+	farcall CopyWhitePal
+	jr .got_previous
+
+.swap_previous
+	xor a
+	assert PREV_PALSTATE == 0
+	ld [wPalState], a
+	call SwapColorPalette
+
+.got_previous
+	pop de
+	pop bc
+	ld a, CURR_PALSTATE
+	ld [wPalState], a
+	push bc
+	call SwapColorPalette
+	pop bc
+	ld a, b
+	farjp CatchUpBGPaletteFade
+
 INCLUDE "data/tileset_palettes.asm"
 
 INCLUDE "data/maps/environment_colors.asm"
--- a/engine/gfx/dma_transfer.asm
+++ b/engine/gfx/dma_transfer.asm
@@ -277,4 +277,8 @@ DI_DelayFrame:
 	jr nc, .loop
 .done
 	pop bc
+	; Don't service the VBlank that elapsed while interrupts were disabled.
+	ldh a, [rIF]
+	res B_IF_VBLANK, a
+	ldh [rIF], a
 	ret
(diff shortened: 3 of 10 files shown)


constants/gfx_constants.asm
```diff
@@ -63,6 +63,15 @@ DEF DEXTILE_FROM_DEXMAP EQU 1 << DEXTILE_FROM_DEXMAP_F
 DEF NUM_FLYFROM_ANIM_OAMS EQU 4 + 9 ; 4 for flymon, 9 for leaves
 DEF NUM_FLYTO_ANIM_OAMS   EQU 4 + 8 ; 4 for flymon, 8 for leaves
 
+; wPaletteSwapFlag uses one nibble each for current and initialized states.
+DEF NUM_PALETTE_SWAP_STATES EQU 4
+DEF PALETTE_SWAP_STATE_MASK EQU (1 << NUM_PALETTE_SWAP_STATES) - 1
+
```
Rangi42: Use of nibbles is a more fundamental design choice that doesn't lend itself to changing these constants at all.
```suggestion
```

macros/scripts/maps.asm
```diff
@@ -225,7 +225,12 @@ MACRO stonetable
 	dw \3 ; script pointer
 ENDM
 
+DEF _NUM_PALETTE_SWAPS = 0
+
 MACRO paletteswap
+	assert _NUM_PALETTE_SWAPS < NUM_PALETTE_SWAP_STATES, \
+		"A palette swap list may have at most {d:NUM_PALETTE_SWAP_STATES} entries"
```
Rangi42:
```suggestion
	; wPaletteSwapFlag uses one nibble each for current and initialized states,
	; so max 4 paletteswaps for the 4 nibble bits.
	assert _NUM_PALETTE_SWAPS < 4, \
		"A paletteswap list may have at most 4 entries"
```

engine/gfx/sprite_palettes.asm
```diff
@@ -79,7 +79,7 @@ CopyLeafGreenToOBPal7:
 ; other colors like purple. (We can assume that if PAL_BG_GREEN is *not* swapped,
 ; then it's still appropriate for use as a leaf/tree color.)
 	ld a, [wPaletteSwapFlag]
-	and a
+	and PALETTE_SWAP_STATE_MASK
```
Rangi42:
```suggestion
	; wPaletteSwapFlag uses one nibble each for current and initialized states.
	and $f
```

engine/gfx/color.asm
```diff
@@ -1057,23 +1058,25 @@ SwapColorPalette::
...
-	add LOW(wBGPals1)
-	ld e, a
-	adc HIGH(wBGPals1)
-	sub e
-	ld d, a
+	ld bc, 1 palettes
+	rst AddNTimes
+	ld d, h
+	ld e, l
+	pop hl
```
Rangi42: I guess it's not a performance bottleneck given the multiple `farcall`s going on in here, but this still saves one byte, and avoids the `push`+`pop`+`rst` overhead on principle.

```suggestion
	; Previous-state palettes become active; current-state palettes are the target.
	; d = [wPalState] == PREV_PALSTATE ? LOW(wBGPals2) : LOW(wBGPals1)
	ld a, [wPalState]
	assert PREV_PALSTATE == 0
	and a
	ld d, LOW(wBGPals2)
	jr z, .got_target
	ld d, LOW(wBGPals1)
.got_target
	; Set `de` to the `b`th palette of `wBGPals1` or `wBGPals2`.
	ld a, b
	add a
	add a
	add a
	add d ; LOW(wBGPals1) or LOW(wBGPals2)
	ld e, a
	assert HIGH(wBGPals1) == HIGH(wBGPals2)
	adc HIGH(wBGPals1)
	sub e
	ld d, a
```

engine/gfx/color.asm
```diff
@@ -1084,6 +1087,71 @@ SwapColorPalette::
...
+; wPaletteSwapFlag's low nibble stores the current state of each entry;
+; its high nibble records which entries have been initialized.
+; Input: `c` PALETTE_SWAP_INSIDE_F set inside the rectangle,
+;        `e` = this entry's state bit.
+; Output: `c` PALETTE_SWAP_CHANGED_F set on first check or a state change.
+	ld a, [wPaletteSwapFlag]
+	ld b, a
+	; Mark this entry initialized, and build its new state in `a`.
+	ld a, e
+	swap a
```
Rangi42: Would it be simpler to just have two bytes, `wPaletteSwapStates` and `wPaletteSwapInits`? That would also raise the limit to 8 `paletteswap`s per map, in other words "more that you'll ever need" rather than "the most you're likely to need".

home/palette_swap.asm
```diff
@@ -1,7 +1,15 @@
 HandlePaletteSwap::
+	ld a, [wPlayerStepFlags]
+	bit PLAYERSTEP_STOP_F, a
+	jr nz, .update
+	farcall CheckPaletteFading
```
Rangi42, line 5: Isn't `homecall` preferable to `farcall` since we're in ROM0?

home/palette_swap.asm
```diff
@@ -66,24 +115,53 @@ InitializeSwappedPalette::
...
+	ld a, d
+	or e
+	jr z, .swapped
+
+	bit PALETTE_SWAP_CHANGED_F, c
+	jr z, .swap_current
+
+	; Rebuild this palette under the fade's previous conditions, then catch it
+	; up to the current fade step toward the same palette's current conditions.
+	farcall CatchUpPaletteSwapFade
```
Rangi42, line 133: I feel uneasy about all these `farcall`s in the home bank, in something that needs to run quickly. How much of this logic could be moved into the relevant ROMX bank so these could be `call`s?

home/palette_swap.asm
```diff
@@ -66,24 +115,53 @@ InitializeSwappedPalette::
...
 	pop hl
+	pop de
 
 	; Continue processing more than one `paletteswap` rectangle.
 	call SwitchToMapScriptsBank
-	jr .loop
+	jr .palette_loop
+
+.get_palette_mask
+; Return a one-hot mask for BG palette ID `b`, preserving `bc`.
```
Rangi42: CS terms are fun but let's keep it simple. :P
```suggestion
; Return `a` = 1 << `b` given BG palette ID `b`, preserving `bc`.
```

home/palette_swap.asm
```diff
@@ -66,24 +115,53 @@ InitializeSwappedPalette::
...
+	push bc
```
Rangi42, line 161: Does `b` actually need preserving? `bc` gets set to 4 right after this.

macros/scripts/events.asm
```diff
@@ -1427,6 +1427,7 @@ ENDM
 
 	const usepaletteswap_command
 MACRO usepaletteswap
+	redef _NUM_PALETTE_SWAPS = 0
```
Rangi42:
```suggestion
	def _NUM_PALETTE_SWAPS = 0
```

macros/scripts/maps.asm
```diff
@@ -225,7 +225,12 @@ MACRO stonetable
 	dw \3 ; script pointer
 ENDM
 
+DEF _NUM_PALETTE_SWAPS = 0
+
 MACRO paletteswap
+	assert _NUM_PALETTE_SWAPS < NUM_PALETTE_SWAP_STATES, \
```
Rangi42: Nitpick, but for the sake of convention, put this sort of increment and assert logic *after* the `db`s which are the main point of the macro, along with the pre-existing `assert`s.

ram/wramx.asm
```diff
@@ -1088,7 +1088,7 @@ wTimeOfDayPal:: db
 wFollowInSync:: db
 
 wPaletteSwapAddress:: dw
-wPaletteSwapFlag:: db
+wPaletteSwapFlag:: db ; low nibble: current states; high nibble: initialized
```
Rangi42: There's a `	ds 1 ; unused` after `wMooMooBerries` which could be used to keep the current states and the initialized markers separate. It's less important that it would raise the limit from 4 to 8, but more important that it would avoid unnecessary masking/`swap`ping.

Rangi42 (changes requested):
Thank you very much!

constants/gfx_constants.asm
vulcandth: Applied in 4a19105afb. The nibble-specific constants and comment are removed; the state tracking now uses separate bytes.

macros/scripts/maps.asm
vulcandth: Addressed via the separate-byte design: the macro now documents one bit per entry and directly asserts the resulting limit of 8 entries.

engine/gfx/sprite_palettes.asm
vulcandth: Addressed via the separate-byte design. This now reads wPaletteSwapStates directly, so no nibble mask is needed.

engine/gfx/color.asm
vulcandth: Applied the suggested address calculation in 4a19105afb, including the LOW/HIGH relationship assertions and comments.

engine/gfx/color.asm
vulcandth: Done. Current states and initialization markers now use separate wPaletteSwapStates and wPaletteSwapInits bytes, raising the enforced limit to 8 and removing the nibble masking/swapping.

home/palette_swap.asm
vulcandth: I kept farcall here because CheckPaletteFading returns its result in A and the Z flag. homecall restores AF after the target returns, which would discard that result and make the following ret nz test the caller's old flags.

home/palette_swap.asm
vulcandth: I reviewed the bank crossings carefully. CatchUpPaletteSwapFade already lives in the color ROMX bank and directly calls the color routines internally, so this site is one crossing per affected palette. The outer parser must repeatedly return to wMapScriptsBank because each paletteswap list is data in the current map-script ROMX bank, which can vary by map; moving the parser wholesale into the color bank would not remove those switches. ROM0 also has only a few spare bytes, so moving the ROMX routines there would require unrelated relocation. I kept the necessary crossings and made UpdatePaletteSwapState cheaper with the separate state/init bytes.

home/palette_swap.asm
vulcandth: Applied the suggested simpler wording in 4a19105afb.

home/palette_swap.asm
vulcandth: The first call site could clobber b, but the second cannot: b is still the BG palette ID passed to CatchUpPaletteSwapFade or SwapColorPalette after the mask check. Since both sites share this helper, it still needs to preserve bc.

macros/scripts/events.asm
vulcandth: Applied in 4a19105afb: this now uses def.

macros/scripts/maps.asm
vulcandth: Applied in 4a19105afb. The count assertion and increment now follow the data and pre-existing bank assertions.

ram/wramx.asm
vulcandth: Applied. The unused byte after wMooMooBerries is now wPaletteSwapInits, the original byte is wPaletteSwapStates, and both are cleared during map initialization. UpdatePaletteSwapState no longer masks or swaps nibbles.

engine/gfx/color.asm
```diff
@@ -1084,6 +1092,74 @@ SwapColorPalette::
...
+	; Store only on first check or if the current state changed.
+.compare
+	cp d
+	jr nz, .changed
+	ld a, [wPaletteSwapInits]
+	and e
+	ret nz
+	ld a, d
+.changed
+	ld [wPaletteSwapStates], a
```
Rangi42: Doing `ld a, d :: ld [wPaletteSwapStates], a` is redundant since we did `ld a, [wPaletteSwapStates] :: ld d, a` earlier, so it sets the same value it already has. A `jr .finish` costs 3 cycles, whereas a `ld a, d :: ld [wPaletteSwapStates], a` costs 5. Saving 2 cycles at the cost of one more ROMX byte is insignificant one way or the other, but I think it's conceptually cleaner to avoid a redundant operation.
```suggestion
	jr .finish
.changed
	ld [wPaletteSwapStates], a
.finish
```

engine/gfx/color.asm
```diff
@@ -1084,6 +1092,74 @@ SwapColorPalette::
...
+	assert PREV_PALSTATE == 0
+	ld [wPalState], a
+	call SwapColorPalette
+
+.got_previous
+	pop de
+	pop bc
+	ld a, CURR_PALSTATE
+	ld [wPalState], a
+	push bc
```
Rangi42: This is redundant, right?
```suggestion
	ld a, CURR_PALSTATE
	ld [wPalState], a
```

engine/gfx/color.asm
```diff
@@ -1084,6 +1092,74 @@ SwapColorPalette::
...
+	push de
+	ld a, [wPalWhiteState]
+	and a
+	jr z, .swap_previous
+	ld hl, wBGPals2
+	ld a, b
+	ld bc, 1 palettes
+	rst AddNTimes
+	ld d, h
+	ld e, l
```
Rangi42: Uses one more ROMX byte but is definitely faster. Assuming you don't need the value in `hl` too.
```suggestion

	; de = wBGPals2 palette b
	ld a, b
	add a
	add a
	add a
	add LOW(wBGPals2)
	ld e, a
	adc HIGH(wBGPals2)
	sub e
	ld d, a

```

engine/gfx/dma_transfer.asm
```diff
@@ -277,4 +277,8 @@ DI_DelayFrame:
 	jr nc, .loop
 .done
 	pop bc
+	; Don't service the VBlank that elapsed while interrupts were disabled.
+	ldh a, [rIF]
+	res B_IF_VBLANK, a
+	ldh [rIF], a
```
Rangi42, line 283: This is a pretty low-level change. What else does it affect? Is it definitely not breaking any non-overworld VBlank handlers? What was buggy before adding this?

engine/gfx/fade.asm
```diff
@@ -27,6 +27,29 @@ OWFadePalettesInit::
...
+	ldh a, [rWBK]
+	push af
+	ld a, BANK(wPalFadeDelayFrames)
+	ldh [rWBK], a
+	ld a, [wPalFadeDelayFrames]
+	ld b, a
+	pop af
+	ldh [rWBK], a
+	ld a, b
+	pop bc
```
Rangi42, line 43: Could this all just use `GetFarWRAMByte`? If not, a brief comment as to why (can't trash `hl`? `push/pop hl` would be more cycle-expensive?) would be appropriate.

engine/gfx/fade.asm
```diff
@@ -381,7 +404,23 @@ FadeColorStep:
 CatchUpObjPaletteFade::
 ; Input: a = OBJ palette index (0-7)
 ; Ensures a newly loaded palette matches the current fade progress.
+	ld hl, wOBPals2
+	jr CatchUpPaletteFade
```
Rangi42, line 411: Remind me why this doesn't need the `ld a, TRUE :: ldh [hCGBPalUpdate], a` behavior; do OBJ palettes update differently?

engine/gfx/fade.asm
```diff
@@ -381,7 +404,23 @@ FadeColorStep:
...
+	ld a, TRUE
+	ldh [hCGBPalUpdate], a
+	ret
+
+CatchUpPaletteFade:
 	ld b, a
+	push bc
+	ld bc, 1 palettes
+	rst AddNTimes
+	pop bc
```
Rangi42: I kind of consider `AddNTimes` a code smell at this point, especially when the registers it uses need preserving. It's only 2 bytes more expensive ROMX-wise to use the optimal `hl = constant + a` or (here) `hl += a` patterns. We know that for `hl += a * (1 palettes)`, the `a * (1 palettes)` quantity (aka `a * 8`) will still fit in a byte, since `1 palettes` is 8 and `a` can only be as high as 7.
```suggestion
	; hl += `a` palettes
	add a
	add a
	add a
	add l
	ld l, a
	adc h
	sub l
	ld h, a
```

engine/gfx/sprite_palettes.asm
```diff
@@ -175,7 +190,18 @@ CopySpritePal::
...
+	; Copy-BG palettes use the corresponding buffer for the requested state.
+	; wBGPals2 is the active, possibly partially faded palette; wBGPals1 is the
+	; destination palette.
+	push af
+	ld a, [wPalState]
+	assert PREV_PALSTATE == 0
+	and a
+	ld hl, wBGPals2
+	jr z, .got_copy_bg_source
 	ld hl, wBGPals1
```
Rangi42: Unnecessary saving of one byte/cycle, but may as well. :P
```suggestion
	assert HIGH(wBGPals2) == HIGH(wBGPals1)
	ld l, LOW(wBGPals1)
```

engine/overworld/warp_connection.asm
```diff
@@ -11,8 +11,9 @@ HandleContinueMap:
 	ld hl, wPaletteSwapAddress
 	ld [hli], a
 	ld [hli], a
-	assert wPaletteSwapAddress + 2 == wPaletteSwapFlag
+	assert wPaletteSwapAddress + 2 == wPaletteSwapStates
 	ld [hl], a
+	ld [wPaletteSwapInits], a
```
Rangi42: Hmm. For the sake of keeping these all together: how about let's change this:

```asm
wFollowInSync:: db

wPaletteSwapAddress:: dw
wPaletteSwapStates:: db

...

wPaletteSwapInits:: db
```

to this:

```asm
wPaletteSwapAddress:: dw
wPaletteSwapStates:: db
wPaletteSwapInits:: db

...

wFollowInSync:: db
```

That avoids shifting anything in-between them, and both palette-swap and follow-in-sync are unreleased master-branch features, so nobody's save file is depending on them. (Follow-in-sync is only used for the Azalea Gym puzzle's Spinarak carts.)

That will allow you to change this here:

```asm
	assert wPaletteSwapAddress + 2 == wPaletteSwapStates
	ld [hl], a
	ld [wPaletteSwapInits], a
```

to this:

```asm
	assert wPaletteSwapAddress + 2 == wPaletteSwapStates
	ld [hli], a
	assert wPaletteSwapStates + 1 == wPaletteSwapInits
	ld [hl], a
```

home/palette_swap.asm
```diff
@@ -17,46 +25,87 @@ InitializeSwappedPalette::
...
 	ret
 
-.loop
-	; If there are no more palette swap rectangles, just return.
+.has_palette_swaps
+	push hl
+	ld a, CURR_PALSTATE
+	ld [wPalState], a
+	farcall CalculateStates
+	ld de, PALETTE_SWAP_INSIDE ; d: changed BG palettes, e: entry state bit
```
Rangi42:
```suggestion
	lb de, 0, PALETTE_SWAP_INSIDE ; d: changed BG palettes, e: entry state bit
```

home/palette_swap.asm
```diff
@@ -66,24 +115,53 @@ InitializeSwappedPalette::
...
+.swapped
 	pop hl
+	pop de
 
 	; Continue processing more than one `paletteswap` rectangle.
 	call SwitchToMapScriptsBank
-	jr .loop
+	jr .palette_loop
+
+.get_palette_mask
```
Rangi42: Since this is only `call`ed, please move it after the `.done` block not before, and give it a nonlocal nonexported label (`GetPaletteMask:`).

macros/scripts/maps.asm
```diff
@@ -225,6 +225,8 @@ MACRO stonetable
 	dw \3 ; script pointer
 ENDM
 
+DEF _NUM_PALETTE_SWAPS = 0
+
```
Rangi42: Any `paletteswap` should be preceded by a `usepaletteswap`, which initializes this already. An undefined constant error is fine if somebody fails to do that.
```suggestion
```

Rangi42 (changes requested):

engine/gfx/dma_transfer.asm
vulcandth: This is technically not related to the PR.. but was fixing the issue of white NPCs showing up from the old map when exiting the pokecom center bug. Basically VBlank:: was running outside of vblank because dma_transfer was running into vblank.

engine/gfx/color.asm
vulcandth: Applied in 3d5bb71478. The first-check path now jumps to the shared finish block without redundantly rewriting wPaletteSwapStates.

engine/gfx/color.asm
vulcandth: Yes. I removed the adjacent pop/push pair. Since the previous-state SwapColorPalette call clobbers bc, that call now preserves bc locally; the optimized copy-white path keeps b intact. The original saved bc is restored only after the current-state swap.

engine/gfx/color.asm
vulcandth: Applied the suggested direct wBGPals2 palette address calculation in 3d5bb71478. hl was not needed afterward.

engine/gfx/fade.asm
vulcandth: It could use GetFarWRAMByte, but this predicate runs every frame on palette-swap maps. Its generic StackCallInWRAMBankA setup is substantially slower than the direct bank switch here, so I kept the fast path and added a comment explaining why.

engine/gfx/fade.asm
vulcandth: Good catch. OBJ palettes do not update differently here. I moved the hCGBPalUpdate request into the shared CatchUpPaletteFade return path, so both BG and OBJ wrappers queue the hardware upload even when catch-up exits early.

engine/gfx/fade.asm
vulcandth: Applied the suggested direct hl += a palettes calculation and removed the AddNTimes register preservation.

engine/gfx/sprite_palettes.asm
vulcandth: Applied in 3d5bb71478 with the shared-high-byte assertion and LOW-only update.

engine/overworld/warp_connection.asm
vulcandth: Applied. The three palette-swap fields are contiguous, wFollowInSync occupies the former initialization byte location, and map initialization now clears all three fields through hl with both layout assertions.

home/palette_swap.asm
vulcandth: Applied in 3d5bb71478: this now uses lb de, 0, PALETTE_SWAP_INSIDE.

home/palette_swap.asm
vulcandth: Applied. GetPaletteMask is now a nonlocal, nonexported label after the .done return block, and both callers use it.

macros/scripts/maps.asm
vulcandth: Applied. The fallback definition is removed, so paletteswap without a preceding usepaletteswap now fails as intended.

engine/gfx/fade.asm
Rangi42:
```suggestion
	lb de, PAL_COLORS, 0
```

Rangi42: Sorry, I know this code isn't new to this PR, but now that we're touching this file again, it could use some extra scrutiny.

engine/gfx/fade.asm
```diff
@@ -412,10 +455,6 @@ CatchUpObjPaletteFade::
 	push bc
```
Rangi42: As far as I can tell, `bc` doesn't need preserving for the `call FadeSinglePaletteStep`. The values of `b`/`c`/`bc` are not needed by this call site at this point in the `CatchUpPaletteFade` logic. And `FadeSinglePaletteStep` does not use `b`/`c`/`bc` as input; it unconditionally writes to them in its initial logic.

I'd appreciate if Codex can take a second look at how the pre-existing `FadeSinglePaletteStep` function works, and consider whether it needs optimizing or removing anything else redundant.

home/palette_swap.asm
```diff
@@ -66,24 +115,41 @@ InitializeSwappedPalette::
...
+
+	; Rebuild this palette under the fade's previous conditions, then catch it
+	; up to the current fade step toward the same palette's current conditions.
+	farcall CatchUpPaletteSwapFade
+	jr .swapped
+
+.swap_current
+	; Always reload the selected palette: CloseSubmenu does not restore
+	; non-default palettes, even when this entry's state has not changed.
+	; TODO: look into a more targeted fix for this.
```
Rangi42, line 139: Not necessary at this point, it's clearly needed due to fade effects.
```suggestion
```

Rangi42 (changes requested):

engine/gfx/fade.asm
vulcandth: Applied in f219bfee3d. The requested initialization is now `lb de, PAL_COLORS, 0`. On the additional scrutiny, the rest of `FadeSinglePaletteStep` was an exact duplicate of `FadePalettesStep.inner_loop`, so the routine now jumps to that existing loop instead of maintaining a second color-fade implementation. This removes 77 duplicated lines and keeps the single-palette and full-fade behavior on the same code path.

engine/gfx/fade.asm
vulcandth: Confirmed and applied in f219bfee3d: `bc` is not live after `FadeSinglePaletteStep`, so the surrounding push/pop pair is removed. I also traced the callee in full and found that its color loop duplicated `FadePalettesStep.inner_loop`; `FadeSinglePaletteStep` now only establishes the four-color count and jumps to that shared implementation.

home/palette_swap.asm
vulcandth: Removed the now-unnecessary catch-up explanation in f219bfee3d.

engine/gfx/fade.asm
Rangi42: Nice catch, thank you!

engine/gfx/fade.asm
```diff
@@ -409,110 +452,27 @@ CatchUpObjPaletteFade::
...
-	ld a, e
-	ld [hli], a
-	ld a, d
-	ld [hli], a
-	pop de
-	dec d
-	jr nz, .single_loop
-	ret
+	lb de, PAL_COLORS, 0
+	jp FadePalettesStep.inner_loop
```
Rangi42: Never use `jp` except in `jp hl`; always use `jmp` to be warned if it could be a more optimal `jr`. Also don't reference others' local labels; rename `FadePalettesStep.inner_loop` to `_DoFadePalettes` (fixing up any locals if necessary). And move `FadeSinglePaletteStep` closer to it, possibly to fallthrough. (*One* of its referers should fallthrough, since one clearly can; and the other should be within `jr` range.)
```suggestion
	jmp FadePalettesStep.inner_loop
```

Rangi42 (changes requested):

engine/gfx/fade.asm
Rangi42: Fixed myself in 1e16655759f6bf489ba916c4d6afe6b42f6e704a.

Rangi42 (approved):
LGTM! Thanks for the persistence, both of you!

PR #1609, opened 2026-08-29
Improve text compression with contextual Huffman coding

vulcandth:
## Summary

Replace the existing order-0 Huffman plus common n-gram text compressor with a three-context Huffman codec designed for the GBC runtime.

The new format recovers **22,260 bytes** in the normal build and leaves two entire 16 KiB switchable ROM banks unused.

## Space savings

| Build | Free space | ROMX banks used |
| --- | ---: | ---: |
| `master` normal build | 12,856 bytes | 127 |
| This PR normal build | 35,116 bytes | 125 |
| Net gain | **22,260 bytes (21.74 KiB)** | **2 fewer banks** |

The clean debug build has 35,008 bytes free.

## Compression format

The encoder selects one of three Huffman trees according to the previously decoded character:

- **Boundary/start:** beginning of a compressed stream, spaces, and line/paragraph controls
- **Vowel:** after an uppercase or lowercase ASCII vowel
- **Other:** after every other character

This captures useful English-language context while keeping the GBC decoder small and predictable. The three trees contain 260 parent nodes total and occupy 520 bytes in ROM0.

Each tree also has an escape leaf. Rare symbols are encoded as the escape code followed by the raw eight-bit character value. The generated maps define this escape representation for all 125 compressible symbols in every context, including combinations absent from the training corpus. This prevents debug-only or newly added text from unexpectedly abandoning compression.

The assembler-side encoder still emits compressed text only when it is smaller than the raw string.

## Dictionary changes

The 67 general-purpose common n-gram entries were removed. Their table, decoder, and charmap overhead outweighed most of their savings once contextual Huffman coding was used.

Five semantic substitutions remain:

- `#` / `Poké`
- `#mon` / `Pokémon`
- `<PLAYER>`
- `<RIVAL>`
- `<TRENDY>`

The last three resolve through their WRAM-backed strings. I also measured replacing `#` and `#mon` with their constituent Huffman-coded source characters, as suggested in review: it reduced free space from 26,231 to 21,155 bytes, a 5,076-byte regression. The two fixed substitutions therefore remain.

## Implementation

- Generate deterministic context-specific codes from the assembled text corpus.
- Assemble three independent decode trees in ROM0.
- Track the previous decoded character in `hCompressedTextBuffer` and select the matching tree at runtime.
- Decode escape leaves by reading the following eight bits literally.
- Preserve context across compressed output written directly to RAM.
- Emit every candidate text token into the training corpus independently of the current table's compression decisions.
- Serialize the `huffman` extraction target so parallel object output cannot interleave and corrupt previous-character context.
- Add a verifier that re-encodes the corpus and decodes it through the tree bytes from the built ROM.

The Huffman macros/constants now live in `constants/huffman_text.asm`; generated codes live in `constants/huffman_text.inc`. `utils/huffman.py` directly contains the generator implementation.

The changed text layout pushed three forward-only branches in `PewterMuseumOfScience1F` beyond their one-byte range. Those instructions are widened to their regular branch forms.

## Debug-build regression fixed

The initial implementation warned in `make -j6 debug` for `%` in the initial-options description. `%` had appeared only after a vowel in the normal training corpus, while the debug string uses it after `0`.

Generated escape codes now cover symbols that have zero observations in a context. The debug string remains compressed instead of falling back to raw text.

## Regenerating the tables

Build artifacts must be tidied before changing build modes. The extraction target intentionally runs serially even when `-j6` is supplied, because previous-character contexts require source-ordered output:

```sh
make tidy
make -j6 huffman > chars.txt
./utils/huffman.py chars.txt constants/huffman_text.inc
```

## Verification

- Two clean corpus extractions match token-for-token: 565,651 tokens in 10,299 streams.
- Regenerating `constants/huffman_text.inc` from either extraction produces identical output.
- Clean debug build: 35,008 bytes free; linked-ROM round trip passes all 565,651 tokens.
- Clean normal build: 35,116 bytes free; linked-ROM round trip passes all 565,651 tokens.
- Python syntax checks and `git diff --check` pass.

Manual checks of the radio shows, Pokédex entries, name entry screens, and PC storage VWF item names are still pending and will be performed by @vulcandth.


constants/charmap.asm
```diff
@@ -1,12 +1,29 @@
 MACRO ctxtmap
```
Rangi42: `ctxtmap` is now redundant and should just use `charmap` directly.

constants/charmap.asm
```diff
@@ -1,12 +1,29 @@
...
 	DEF x = \2
-	DEF ___huffman_data_{02X:x} EQU %\3
-	DEF ___huffman_length_{02X:x} EQU STRLEN("\3")
-	DEF ___huffman_char_\3 EQUS "\1"
 	charmap \1, x
 ENDM
 
-; Huffman text compression (see data/text/compressed_text.asm and home/text.asm)
+MACRO huffmap
+	DEF x = CHARVAL(\2, 0)
```
Rangi42: From the rgbasm docs: "If *idx* is not specified, *char* must have a single value, which is returned." This is simpler to read, and correctly fails to build if `\2` is a multi-value charmap, since the whole system assumes one byte value per char.
```suggestion
	DEF x = CHARVAL(\2)
```

constants/charmap.asm
```diff
@@ -1,12 +1,29 @@
...
+	DEF ___huffman_data_\1_{02X:x} EQU %\3
+	DEF ___huffman_length_\1_{02X:x} EQU STRLEN("\3")
+ENDM
+
+MACRO huffleaf
+	DEF ___huffman_escape_\1_\2 = 1
+ENDM
+
+; Contextual Huffman text compression
```
Rangi42: The Huffman-related macros and constants should go in their own `constants/huffman_text.asm` file, since none of them are needed any more for defining the `charmap`s in this file.

constants/charmap.asm
```diff
@@ -245,92 +244,25 @@ DEF FIRST_REGULAR_TEXT_CHAR EQU $7f
...
-DEF NGRAMS_START EQU $0a
-
 	newcharmap compressing, no_ngrams
 
-	ctxtmap "#",        $4d, 10000111010
-	ctxtmap "#mon",     $4e, 01010000
+DEF NGRAMS_START EQU $0a ; first value that implicitly starts text
+DEF TEXT_SUBSTITUTIONS_START EQU $4d
+	ctxtmap "#",        $4d
+	ctxtmap "#mon",     $4e
```
Rangi42: Do either of these save space any more? Would it be simpler and save more space to treat `"#"` as a placeholder for four `"Poké"` characters at the *source code* level, so once the ROM is assembled, it gets Huffman encoded as those four separate chars?

constants/text_huffman_contexts.asm
Rangi42, line 1: How about `constants/huffman_text.inc`, to mark this as a generated file. It would be `INCLUDE`d in `constants/huffman_text.asm` with the relevant macros and constants currently in `constants/charmap.asm`.

constants/charmap.asm
```diff
@@ -245,92 +244,25 @@ DEF FIRST_REGULAR_TEXT_CHAR EQU $7f
...
 
 
-DEF NGRAMS_START EQU $0a
-
 	newcharmap compressing, no_ngrams
 
-	ctxtmap "#",        $4d, 10000111010
-	ctxtmap "#mon",     $4e, 01010000
+DEF NGRAMS_START EQU $0a ; first value that implicitly starts text
+DEF TEXT_SUBSTITUTIONS_START EQU $4d
```
Rangi42: From the code further down, it looks like `NGRAMS_START` is now unused, because all uses of it were replaced with `TEXT_SUBSTITUTIONS_START`. It should have just changed the value of `NGRAMS_START` from `$0a` to `$4d`, and left the usage sites alone.

home/text.asm
```diff
@@ -779,6 +779,10 @@ DecompressString::
 
 	inc hl ; skip "<CTXT>"
 
+	; Start the contextual decoder at a word boundary.
```
Rangi42: Nitpicky, but the surrounding comments have this style.
```suggestion
	; start the contextual decoder at a word boundary
```

home/text.asm
```diff
@@ -874,12 +878,15 @@ DecompressStringToRAM::
 	inc hl ; skip "<CTXT>"
 
 .do_decompression
+	ld a, ' '
+	ldh [hCompressedTextBuffer], a
```
Rangi42, line 883:
```suggestion
	; start the contextual decoder at a word boundary
	ld a, ' '
	ldh [hCompressedTextBuffer], a
```

home/text.asm
```diff
@@ -908,34 +915,98 @@ DecompressStringToRAM::
...
+	; Select a tree from the previous decoded character.
+	ldh a, [hCompressedTextBuffer]
+	cp ' '
+	jr z, .boundary
+	cp '<LNBRK>'
+	jr c, .check_vowel
+	cp '<PARA>' + 1
+	jr c, .boundary
+.check_vowel
+	and $df ; normalize lowercase letters to uppercase
```
Rangi42:
```suggestion
	and ~('A' ^ 'a') ; normalize lowercase letters to uppercase
```

macros/scripts/text.asm
```diff
@@ -180,9 +185,32 @@ ENDM
...
+			DEF _huffman_data = ___huffman_data_v_{02X:_chr}
+			DEF _huffman_length = ___huffman_length_v_{02X:_chr}
+		endc
+	else
+		if DEF(___huffman_data_o_{02X:_chr}) && DEF(___huffman_length_o_{02X:_chr})
+			DEF _huffman_defined = 1
+			DEF _huffman_data = ___huffman_data_o_{02X:_chr}
+			DEF _huffman_length = ___huffman_length_o_{02X:_chr}
+		endc
+	endc
```
Rangi42: This is repeating the same stuff three times, but manually mapping these 0/1/2 sentinel values to b/v/o sentinel characters. It should just reuse the same sentinel values in names, so it can do the actual logic once:
```asm
	if DEF(___huffman_data_{_ctxt_context}_{02X:_chr}) && DEF(___huffman_length_{_ctxt_context}_{02X:_chr})
		DEF _huffman_defined = 1
		DEF _huffman_data = ___huffman_data_{_ctxt_context}_{02X:_chr}
		DEF _huffman_length = ___huffman_length_{_ctxt_context}_{02X:_chr}
	endc
```

maps/PewterMuseumOfScience1F.asm
Rangi42: Why was this changed? Did some text *grow* to make them no longer in `fwd` range?

utils/huffman.py
Rangi42, line 1: We don't need a "compatibility entry point"; just overwrite the previous contents of `utils/huffman.py` with the new logic.

macros/scripts/text.asm
```diff
@@ -235,6 +260,14 @@ MACRO _dchr
...
 			endc
 		endr
+		if _chr == ' ' || ('<LNBRK>' <= _chr && _chr <= '<PARA>')
+			DEF _ctxt_context = CTXT_CONTEXT_BOUNDARY
+		elif _chr == 'A' || _chr == 'E' || _chr == 'I' || _chr == 'O' || _chr == 'U' \
+			|| _chr == 'a' || _chr == 'e' || _chr == 'i' || _chr == 'o' || _chr == 'u'
+			DEF _ctxt_context = CTXT_CONTEXT_VOWEL
+		else
+			DEF _ctxt_context = CTXT_CONTEXT_OTHER
+		endc
```
Rangi42, line 271:
```suggestion
		if _chr == ' ' || ('<LNBRK>' <= _chr && _chr <= '<PARA>')
			REDEF _ctxt_context EQUS "b"
		elif _chr == 'A' || _chr == 'E' || _chr == 'I' || _chr == 'O' || _chr == 'U' \
			|| _chr == 'a' || _chr == 'e' || _chr == 'i' || _chr == 'o' || _chr == 'u'
			REDEF _ctxt_context EQUS "v"
		else
			REDEF _ctxt_context EQUS "o"
		endc
```

Rangi42 (changes requested):
Wow, thank you, this is startling and valuable work. I left some preliminary comments, and once they're addressed, will look further into how the new system works.

Rangi42:
> No emulator gameplay test was available in the local environment.

@vulcandth I assume you did at least cursory gameplay checking to verify this works? Did you check particular texts which have had issues before or would exercise edge cases, like the radio shows, the Pokédex entries, the name entry screens, the PC storage display of VWF item names?

home/text.asm
vulcandth: Applied exactly.

constants/charmap.asm
vulcandth: Done in 8691bc0148: removed `ctxtmap` and converted the declarations to plain `charmap`.

constants/charmap.asm
vulcandth: I measured this before deciding. Expanding `#` and `#mon` into separately Huffman-coded characters reduced free space from 26,231 to 21,155 bytes, a 5,076-byte regression, and pushed two nearby branches out of JR range. The fixed substitutions therefore still save substantial space, so I retained them and documented the measurement in the PR.

constants/charmap.asm
vulcandth: Done: both Huffman map macros now use `CHARVAL(\\2)` with no index, and the tree builder does the same for generated leaf names.

constants/charmap.asm
vulcandth: Addressed with one necessary split. `NGRAMS_START` is now `$4d`, and all n-gram/substitution indexing uses it. The old `$0a` value also sized and dispatched the ten-entry `TextCommands` table, so that independent role is now named `NUM_TEXT_COMMANDS`; using `$4d` there would incorrectly require 77 command entries.

home/text.asm
vulcandth: Applied exactly.

utils/huffman.py
vulcandth: Done: `utils/huffman.py` now contains the generator directly, and `utils/huffman_context.py` was removed.

macros/scripts/text.asm
vulcandth: Done: `_ctxt_context` now directly holds the `b`/`v`/`o` name, and the Huffman lookup logic appears only once.

constants/text_huffman_contexts.asm
vulcandth: Done: the generated file is now `constants/huffman_text.inc`, included by the hand-written `constants/huffman_text.asm`.

home/text.asm
vulcandth: Applied exactly.

macros/scripts/text.asm
vulcandth: Done: the context updates now directly `REDEF` the `b`/`v`/`o` sentinel names as suggested.

constants/charmap.asm
vulcandth: Done: the macros, node constants, and generated include now live in `constants/huffman_text.asm` and `constants/huffman_text.inc`.

vulcandth:
I had not manually exercised those specific paths, so I do not want to overstate the testing. The linked-ROM verifier now covers every 565,651 emitted text token in both normal and debug builds, including control characters and RAM decompression streams, but @vulcandth will manually check the radio shows, Pokédex entries, name entry screens, and PC storage VWF item-name display. I updated the PR description to make that pending manual pass explicit.

constants/text_constants.asm
```diff
@@ -7,6 +7,8 @@ DEF MOVE_NAME_LENGTH   EQU 13
 DEF ITEM_NAME_LENGTH   EQU 13
 DEF TRAINER_CLASS_NAME_LENGTH EQU 13
 
+DEF NUM_TEXT_COMMANDS EQU $0a
```
Rangi42: Since this constant's value is tied to the charmaps, it belongs in charmap.asm:

```asm
	charmap "<PLURAL>", $09
DEF NUM_TEXT_COMMANDS EQU $0a
```

constants/charmap.asm
```diff
@@ -39,18 +17,18 @@ assert FIRST_SHIFTED_LEAF_CHAR_ID < LAST_SHIFTED_LEAF_CHAR_ID && LAST_SHIFTED_LE
 
 DEF SPECIALS_START EQU $52
```
Rangi42, line 24: Just to clarify: since characters $0a-$4c are no longer used for ngrams, they're totally free now, right? We could add more text commands starting from $0a upward, or add more ngrams (like `#` and `#mon`) starting from $4c downward, or shift the values of the few ngram characters downward to add more "special" characters (like `<DONE>`, `@`, `<PROMPT>`, all those `SPECIALS_START` ones). IMO the `; n-grams: $0a - $51 (defined below)` comment right above here should be updated to reflect this, since ngrams only take $4d-$51 now.

Rangi42 (changes requested):

Rangi42:
Text looks correct to me. That includes Pokédex entries, radio shows (in the Pokégear and the overworld), ability descriptions, VWF (for item names in the PC and ability names in battle), Mail (reading and writing), name entry screens, in-battle text, various overworld dialog, the Start menu, the initial and in-game option descriptions, the remaining N-grams (`<RIVAL>`, `<PLAYER>`, `<TRENDY>`, `#`, `#mon`)...

I don't see any lag/delay in rendering text either.

The stuff moved out of ROM0 to make room (`FastPrintNum` in the Pokédex, `CheckVBA` warning, `PrintWinLossText` for battles you can win or lose) still works.

Rangi42:
Unfortunately this 100% fills up ROM0, but we can move more things out of there when necessary.

Merging for now, so any more subtle bugs can be found and fixed.

PR #1610, opened 2026-08-29
Compress music streams with Huffman coding

vulcandth:
## Summary

This PR Huffman-compresses music command streams and decodes them incrementally while the audio engine plays them. It is now rebased directly onto `master`, including the merged contextual text-compression work from #1609.

Music headers and relocatable control-flow pointers remain byte-aligned and raw. SFX and cry streams continue through the original uncompressed path.

The selected capped byte-Huffman format uses:

- 127 directly coded byte values;
- one escape leaf followed by an 8-bit literal for remaining values;
- a 254-byte branch tree plus a 127-byte leaf map (381 bytes total);
- a one-byte bit cache and one-byte remaining-bit count per channel;
- an explicit compression byte in each top-level four-byte music-table entry, followed by the full `dba` pointer.

The explicit flag preserves the complete MBC5 bank range; unlike the earlier bank-bit marker experiment, it imposes no bank-below-`$80` limit on a future 4 MiB ROM.

## Savings

Measured from clean normal builds against current `master` (`5a866cafff`):

| Build | Free ROM space | Highest occupied ROMX bank |
|---|---:|---:|
| `master` | 34,971 bytes (1.67%) | `$7d` |
| This PR | 59,502 bytes (2.84%) | `$7c` |
| **Net improvement** | **24,531 bytes (23.96 KiB)** | **1 fewer occupied bank** |

The uncompressed corpus contains 154,829 bytes of channel streams plus 2,094 bytes of raw headers. The selected model reduces stream data by 25,501 bytes, or 25,120 bytes after its 381-byte decode table. The final 24,531-byte net figure includes decoder code, four-byte top-level pointer entries, alignment commands/padding, and all runtime support.

The clean debug build has 59,394 bytes free.

## Build-time encoding and labels

The audio-script macros route music command bytes through a streamed bit packer. Compression begins at each channel stream. Calls, jumps, loops, conditional jumps, channel pointers, and restart-channel operands stay aligned and are emitted as relocatable raw 16-bit addresses.

Every one of the 2,930 original music source labels is retained exactly, including exported `::` labels and local labels such as `.loop2`. A no-argument, column-1 `music_header` or `music_label` obtains that ordinary label's context through `\#` and creates a separate internal byte-aligned symbol for compressed runtime pointers. This preserves familiar navigation and Crystal Tracker context while supporting local branch/call targets correctly. Crystal Tracker will still need compressed-stream decoding support.

The all-song VibeEmu pass exposed an alignment edge case: if one file ended with a non-byte-aligned command, its final padding byte could be emitted after the next song's ordinary source label. Top-level compressed pointers now target an internal header label emitted after `_music_align`, while the original label remains present. This fixed `MUSIC_CAPTURE` starting channels 1 and 5 instead of 1–3.

Sequential fall-through across a source label requires the same bit reset as a taken pointer. The assembler emits the previously unused `$f8` command before such boundaries; fall-through executes it to discard padding, while branches target the internal symbol after the marker.

`constants/music_huffman.asm` contains the generated model. `make music-huffman` emits a deterministic source corpus, and `utils/music_huffman.py` regenerates the table.

## Runtime decoding

`GetMusicByte` checks `SOUND_COMPRESSED`, a previously free bit in each channel's existing Flags byte. Compressed music walks the compact tree one buffered bit at a time until it reaches a direct leaf or the raw-byte escape. SFX and cries do not set the flag and use the existing byte path.

The top-level music table stores `compression flag, bank, address low, address high`. This costs one byte per song entry (193 bytes here) but keeps the full bank byte intact.

The two new cache bytes are appended to `channel_struct`, so established member offsets within each channel remain unchanged. The eight channel structures grow by 16 bytes in total, and one global byte carries the table flag during initialization. All 17 bytes come from the existing unused WRAM0 reservation; alignment padding was adjusted so the layout still links exactly.

## VibeEmu profiling

The complete default playback path of all 192 songs (everything except `MUSIC_NONE`) was compared with an uncompressed ROM built from the exact `master` parent. The profiler executed 2,125,180 paired `_UpdateSound` calls, covering one complete command loop for every active channel or the end of each jingle, and checked established channel state after every call.

Weighted across the catalog, decoding adds 297.22 VibeEmu dot/timing cycles per update: +9.75% relative to the audio routine and 0.423% of a 70,224-cycle frame. The worst average is `MUSIC_MEWTWO_BATTLE_STADIUM`: +838.41 cycles, +24.63%, or 1.194% of a frame per update.

The important tail cases are:

- `MUSIC_POWER_PLANT_XY`, used in Rock Tunnel and Magnet Tunnel: first update 76,938 cycles (109.56% of a frame), 56,030 above master. This is the only default song call over one full frame. Its worst later call is 49,920 cycles (71.09% of a frame).
- `MUSIC_GYM_LEADER_BATTLE_SWSH`, used by Piers and Larry: worst neutral steady-state call 63,450 cycles (90.35% of a frame), 44,234 above master, approximately 43.0 seconds after playback begins.
- The actual final-Pokémon restart path for the same Piers/Larry track is the overall maximum: first update 77,316 cycles (110.10% of a frame), 55,758 above master.

The top candidates and final-Pokémon case reproduced identically across three fresh emulator boots. No steady-state audio-only call exceeds a frame, but a 90.35% call leaves little time for game logic, so these are explicit manual-test candidates rather than being dismissed by the low mean. Full methodology and rankings are in the profiling PR comment.

No CPU-driven optimization is included merely because the profile found a tail cost; this PR reports the tradeoff for review.

## Tooling

- `utils/music_compression_benchmark.py` extracts the uncompressed corpus and compares byte Huffman, capped Huffman, PackBits RLE, and a zlib ceiling.
- `utils/music_huffman.py` builds the capped tree/table used by RGBDS.
- `utils/verify_music_huffman.py` decodes the linked ROM byte-for-byte against the assembler corpus and validates raw pointer targets.

## Validation

- Clean normal build: `make tidy && make -j6` — 59,502 bytes free.
- Clean debug build: `make tidy && make -j6 debug` — 59,394 bytes free.
- Linked-ROM verifier, normal and debug: 148,748 logical music bytes across 5,511 non-empty restart blocks and 3,832 raw control-flow pointers.
- Exact comparison confirms all 2,930 pre-compression source labels are retained.
- VibeEmu differential run: all 192 songs; 2,125,180 paired `_UpdateSound` calls; 187 looped tracks and five ending jingles; matching established channel state after every call.
- Fresh-boot reproduction: Power Plant, neutral Piers/Larry, and final-Pokémon Piers/Larry cases reproduced exactly three times.
- `git diff --check` passes.

Manual listening and visual artifact testing is intentionally left to @vulcandth's follow-up pass.

## Tradeoffs

- Music-byte fetching costs CPU time for tree traversal; both the average and material tail costs are reported above.
- Crystal Tracker retains the ordinary labels/context but must learn the finalized compressed format before it can interpret command bytes.
- SFX and cries remain uncompressed to preserve their behavior and latency.


Rangi42 (changes requested):
This is an impressive PR, both for the creative idea and for the space savings.

Unfortunately it would break compatibility with [Crystal Tracker](https://github.com/dannye/crystal-tracker/), but in theory we could feature-request/PR support for this format, if/when it's finalized and merged. At least the only differences in the music files themselves are replacing labels with `music_header` and `music_label` macros.

Speaking of which, for the sake of easily skimming music files to see where their labels are, please dedent those `music_header` and `music_label` macros so they start at column 1. Remember, it's okay now for rgbasm macros to do that.

> Compressed music banks must remain below bank $80; the pointer macro asserts this at build time.

Given that we eventually plan to move to a 4 MB ROM no matter how much space gets saved by compression (since 10-bit moves and items and Gen 3+ mons will inevitably require it), I don't think this is an acceptable design limitation. It's annoying to manually keep music within certain banks, because rgbasm/rgblink have no feature to automatically do that. Please choose another encoding scheme. AIUI, the "compressed stream marker" just occurs one time per channel per song, so using a whole extra byte for it would spend on the order of 256x4=1024 bytes max.

macros/ram.asm
Rangi42, line 1: Just noting that this will shift a lot of WRAM0, which the save patch will need to account for.

vulcandth: Addressed in 40dd4ebb2b. The two Huffman cache bytes now live at the end of `channel_struct`, so every established member keeps its original per-channel offset. Each channel is still two bytes longer, so later channel bases and subsequent WRAM move; `wMusicCompressed` adds one transient initialization byte, with alignment padding adjusted afterward. All 17 bytes come from the unused WRAM0 reservation. The save patch will still need to account for those base/address shifts, and the PR description now calls that out explicitly.

vulcandth:
Review update pushed in 40dd4ebb2b (rebased onto #1609 at 8691bc0148):

- Restored all 2,930 original music labels exactly, including `::` exports and local labels. `music_header`/`music_label` are now no-argument column-1 macros that obtain the ordinary label context with `\#`; compressed control flow targets separate aligned restart symbols. This restores the source shape useful to Crystal Tracker, although Tracker will still need decoder support for the finalized byte format.
- Local labels are supported. The rebuild exposed three existing Prism `.loop2`/`.loop3` targets that the prior conversion had missed; they are now verified restart boundaries.
- Removed the bank-bit marker and bank-`$80` limit. Each of the 193 top-level entries is now an explicit compression byte plus a full `dba`, so future 4 MiB placement is unrestricted.
- Moved per-channel cache fields to the end of the struct to preserve all established member offsets; documented the remaining channel-base/subsequent-WRAM shifts for the save patch.
- Reprofiled complete `_UpdateSound` calls with VibeEmu against the current uncompressed #1609 parent: 20,000 synchronized calls, all established channel state matched. Mean time rose from 3,154.36 to 3,450.81 timing cycles (+9.40%), which adds 0.422% of a 70,224-cycle frame on average. Median overhead was zero; p95/p99 were 2,230/5,416 cycles. The maximum observed added cost was 43,956 cycles, with the complete compressed call reaching 59,768 cycles (85.1% of a frame). The average is modest, but the rare transition spike is material and is documented candidly in the PR body.
- Clean normal/debug builds have 59,647/59,539 bytes free. Net savings versus the current #1609 parent are 24,531 bytes. Both linked ROMs verify all 148,748 music bytes across 5,511 blocks and 3,832 control-flow pointers.

Per the latest direction, manual in-game edge-case testing remains for @vulcandth’s follow-up pass.

vulcandth:
## Complete VibeEmu music profile after rebasing onto `master`

I reran the profile against an exact uncompressed build of current `master` (`5a866cafff`) after rebasing this branch. This pass covers **all 192 songs** (IDs `$01`–`$c0`, including the six entries marked unused/music-player-only).

### Methodology

- Emulator: local VibeEmu core, not PyBoy.
- Both ROMs boot normally to the audio engine. For each ID, the harness calls the real `PlayMusic` path with `MUSIC_NONE`, then the requested song, and invokes the real `_UpdateSound` routine in paired compressed/uncompressed emulators.
- Timing is taken from VibeEmu's dot/timing-cycle counter. One GBC frame is 70,224 dot cycles.
- A song is sampled until every active music channel has either ended or repeated its complete sequencer state. The loop signature keeps addresses, return addresses, flags, conditions, notes, envelopes, vibrato, pitch, tempo, and the rest of the state that determines playback. It ignores only the fractional note-duration remainder and inactive loop counter; otherwise some musical loops take 256 repetitions to recur solely because sub-frame tempo error is carried forward.
- After **every paired call**, I compare all established channel state. Compressed source pointers and the two new bit-cache bytes are necessarily representation-specific and are excluded. Any musical/control-state divergence aborts the run.
- This produced **2,125,180 paired `_UpdateSound` calls**: 187 tracks reached a loop and five jingles ended. All calls matched the uncompressed build.
- I separately ran all four conditions of the variable Sword/Shield gym track (`default`, `advantage`, `disadvantage`, and `final mon`) because it became the worst steady-state candidate in the default sweep.
- The reported “steady” maximum excludes call 1, which separates ongoing playback from song-start decoding.

This is intentionally an isolated audio-routine profile, not a simulated gameplay benchmark. It gives an exact audio CPU budget and reproducible points for manual listening/visual testing, but it does not claim that VibeEmu observed a rendered dropped frame in a full scene.

### Catalog-wide result

Weighted by all 2,125,180 updates:

| Metric | Uncompressed `master` | Compressed | Difference |
|---|---:|---:|---:|
| Mean `_UpdateSound` cost | 3,048.87 cycles | 3,346.08 cycles | **+297.22 / +9.75%** |
| Mean share of one frame | 4.341% | 4.765% | **+0.423% of a frame** |

The median added cost is zero for the per-song distributions: most updates consume no new music command byte. Cost arrives in bursts on note/command boundaries.

Tracks with the largest average decoding cost:

| Track | Mean added cycles | Relative audio cost | Added frame share |
|---|---:|---:|---:|
| `MUSIC_MEWTWO_BATTLE_STADIUM` | 838.41 | +24.63% | +1.194% |
| `MUSIC_KIMONO_ENCOUNTER` | 667.31 | +23.95% | +0.950% |
| `MUSIC_BATTLE_FACTORY_RSE` | 571.35 | +16.98% | +0.814% |
| `MUSIC_CHAMPION_BATTLE_B2W2` | 568.15 | +16.62% | +0.809% |
| `MUSIC_ELITE_FOUR_BATTLE_SM` | 566.98 | +16.77% | +0.807% |

The largest per-song 99th-percentile added cost is `MUSIC_LASS_ENCOUNTER`, 13,090 cycles (18.64% of a frame). This is another way of showing that the tail is more important than the average.

### Maximum default-path spike: Power Plant XY

`MUSIC_POWER_PLANT_XY` is the maximum in the 192-song default sweep, and it **is used in-game**, not only by the music player:

- Rock Tunnel 1F, B1F, and 2F
- Magnet Tunnel Inside

Its first `_UpdateSound` call is:

| Build | Cycles | Frame share |
|---|---:|---:|
| `master` | 20,908 | 29.77% |
| compressed | **76,938** | **109.56%** |
| added | **56,030** | **79.79%** |

That audio call alone is 6,714 cycles (9.56% of a frame) over the frame budget. It is the **only call over one frame in the default sweep**. Its worst post-start call occurs at update 3,938 (about 65.93 seconds): 49,920 cycles total (71.09% of a frame), 32,578 cycles above `master`.

This result was identical across three fresh VibeEmu boots.

### Overall in-game maximum and worst steady state: Piers/Larry

`MUSIC_GYM_LEADER_BATTLE_SWSH` is used for:

- Piers in Goldenrod Underground (initial fight and rematch)
- Larry in Shamouti Tourist Center

It has the worst default/neutral steady-state call:

| Build | Cycles | Frame share |
|---|---:|---:|
| `master` | 19,216 | 27.36% |
| compressed | **63,450** | **90.35%** |
| added | **44,234** | **62.99%** |

That occurs at update 2,569, approximately **43.01 seconds after the music starts**, at the transition into the neutral loop. The isolated audio routine leaves only 6,774 cycles (9.65% of a frame) for all other game work on that frame. It does not exceed a frame by itself, but it is the strongest recurring manual-test candidate.

The actual final-Pokémon path is more severe. `ResetVariableBattleMusicCondition` stops and restarts this track, then sets `COND_FINAL_MON`; reproducing that same order gives the overall maximum observed call:

| Build | Cycles | Frame share |
|---|---:|---:|
| `master` | 21,558 | 30.70% |
| compressed | **77,316** | **110.10%** |
| added | **55,758** | **79.40%** |

This is a one-time spike on the first audio update after the final-Pokémon restart and is 7,092 cycles over one frame. It reproduced identically on three fresh boots.

All four track conditions:

| Condition | Mean added cycles | Mean added frame share | Maximum after call 1 |
|---|---:|---:|---:|
| neutral/default | 460.62 | 0.656% | **63,450 cycles / 90.35% frame** |
| advantage | 418.06 | 0.595% | 58,892 / 83.86% |
| disadvantage | 427.65 | 0.609% | 61,802 / 88.01% |
| final Pokémon | 289.25 | 0.412% | 35,786 / 50.96% |

The final condition's problem is its one-time restart spike, while neutral is the worst ongoing path.

### Suggested manual reproduction

For Sylvie's listen/look-for-frame-artifacts pass:

1. Fight Piers in Goldenrod Underground or Larry in Shamouti Tourist Center.
2. For the repeatable neutral candidate, keep the music in its neutral condition and watch/listen around **43 seconds after battle music begins**. Battle state changes can switch the conditional arrangement, so a quiet/neutral wait is the closest match to this trace.
3. Watch the exact transition to their final Pokémon. The engine restarts the song there; the **first audio update after that restart** is the 77,316-cycle overall maximum.
4. Separately enter Rock Tunnel or Magnet Tunnel and watch the first frame of `MUSIC_POWER_PLANT_XY`; its first default update is 76,938 cycles. Its worst later point is around **65.9 seconds**.

### Other high steady-state candidates

| Track | Added cycles | Complete compressed call | Frame share |
|---|---:|---:|---:|
| `MUSIC_CELESTIAL_TOWER_BW` | 41,276 | 58,160 | 82.82% |
| `MUSIC_MARINE_TUBE_B2W2` | 39,316 | 56,946 | 81.09% |
| `MUSIC_ROAD_TO_REVERSAL_MOUNTAIN_B2W2` | 38,968 | 56,282 | 80.15% |
| `MUSIC_ROUTE_12_BW` | 36,404 | 53,394 | 76.03% |
| `MUSIC_SAFARI_ZONE_GATE_HGSS` | 33,906 | 48,220 | 68.67% |

These are all used by in-game maps. No steady-state audio-only call in the complete default sweep exceeds 70,224 cycles.

### Additional finding fixed during this pass

The initial all-song run stopped at `MUSIC_CAPTURE`: compressed playback selected channels 1 and 5 instead of master’s channels 1–3. The previous song ended with an unaligned compressed command, and its padding byte was emitted after `Music_SuccessfulCapture` but before its raw header. The pointer table therefore addressed padding as if it were the header.

The pushed fix gives each compressed song an internal header label after `_music_align` and points the top-level table there. Ordinary original labels remain intact for source navigation/tracker context. After the fix, `MUSIC_CAPTURE` and all remaining songs match `master`, and the ROM saving remains **24,531 bytes**.

My conclusion: the average cost is small, but the two >1-frame transition spikes and the 90.35% neutral Piers/Larry call are material enough that manual in-game visual/audio checks are warranted before merging. I have not changed the decoder based solely on these measurements.

Rangi42:
Human testing by listening to the audio won't be viable until #1611 is fixed.

PR #1612, opened 2026-08-30
Add fading campfire and aquarium object glows

vulcandth:
## Summary

Add collision-driven lighting for the player and NPC object sprites near campfires and aquariums.

Objects standing on `COLL_CAMPFIRE_GLOW` or `COLL_AQUARIUM_GLOW` receive a corresponding palette treatment. Entering and leaving either collision initiates a normal overworld palette fade instead of changing the object palette instantly.

## Glow behavior

- Campfire glow selects the object's daytime overworld palette. This makes an object near a campfire appear lit even at night, during darkness, or under overcast palette states.
- Aquarium glow adds a cool BGR555 adjustment of `+2 red`, `+5 green`, and `+9 blue`, with each channel clamped to 31.
- Aquarium tinting only modifies colors 0 through 2. Color 3, conventionally the black outline color, is left unchanged.
- The behavior applies to the player and every active NPC object struct.
- Despawned object slots are cleared without starting a pointless visible exit fade.

## Fade integration

Each object tracks both its previous and current glow type. A collision transition therefore constructs two palette endpoints:

- the active endpoint uses the previous glow state;
- the destination endpoint uses the current glow state.

If no palette fade is active, the change saves the current palette states and starts the standard 15-step overworld fade. If a time-of-day, weather, lightning, map-transition, palette-swap, or other overworld fade is already active, it is preserved rather than restarted. The existing dynamic object-palette catch-up routine rebuilds both endpoints and advances the newly loaded palette to the current fade step.

When the fade finishes, temporary previous-glow state is cleared and dynamic palettes are rescanned so transition-only slots can be deduplicated.

## Shared dynamic palettes

Two objects that request the same base palette do not both become illuminated when only one enters a glow collision.

Dynamic object palette identity now consists of:

`base palette + palette type + previous glow + current glow`

For example, if the player and an NPC normally share blue and only the player enters an aquarium collision, the NPC retains the normal-blue slot while the player is assigned a separate normal-blue to aquarium-blue transition slot. The player's `OBJECT_PALETTE` bits are updated to point at the new slot. Objects with identical base palettes and identical glow endpoints may still safely share a slot.

The existing hardware limit of eight object-palette slots remains. If a scene requires more than eight unique composite object palettes simultaneously, the existing dynamic-palette slot-7 fallback still applies.

## Implementation notes

- Check object collision state after player and NPC stepping each frame.
- Derive collision from coordinates during full dynamic-palette scans so newly spawned or reused object structs start with the correct glow.
- Cache current and previous glow metadata alongside each loaded dynamic palette slot.
- Reuse the existing fade-aware `CopyObjectSpritePalHandler` path for endpoint selection and fade catch-up.
- Clear glow metadata alongside the existing saved dynamic-palette state during map/weather reset paths.
- Use the previously unused WRAM area for glow cache and transition state, leaving two bytes unused.
- Use `jmp` for long unconditional branches so the macro continues to flag branches that could be shortened to `jr`.

## Testing

- Ran `python3 utils/optimize.py`: `Found 0 instances.`
- `git diff --check`
- Clean normal build: `make tidy && make -j2`
- Booted the resulting ROM for 600 frames with PyBoy.
- Monochrome assembly reaches the same two pre-existing unrelated assertions:
  - `CycleJudgeMachinePalette`: expected 64 bytes, got 128 bytes
  - `SingleObjectPals`: expected 176 bytes, got 144 bytes
- Ran `make tidy` before every build-type switch and restored a successful normal build afterward.


constants/gfx_constants.asm
```diff
@@ -63,6 +63,14 @@ DEF DEXTILE_FROM_DEXMAP EQU 1 << DEXTILE_FROM_DEXMAP_F
 DEF NUM_FLYFROM_ANIM_OAMS EQU 4 + 9 ; 4 for flymon, 9 for leaves
 DEF NUM_FLYTO_ANIM_OAMS   EQU 4 + 8 ; 4 for flymon, 8 for leaves
 
+; Dynamic overworld object palette glow types.
+DEF OBJ_GLOW_NONE      EQU 0
+DEF OBJ_GLOW_CAMPFIRE  EQU 1
+DEF OBJ_GLOW_AQUARIUM  EQU 2
+DEF NUM_OBJ_GLOW_TYPES EQU 3
```
Rangi42:
```suggestion
	const_def
	const OBJ_GLOW_NONE
	const OBJ_GLOW_CAMPFIRE
	const OBJ_GLOW_AQUARIUM
DEF NUM_OBJ_GLOW_TYPES EQU const_value
```

engine/gfx/dynamic_pals.asm
```diff
@@ -56,6 +64,14 @@ CheckForUsedObjPals::
 	bit DISABLE_DYN_PAL_F, [hl]
 	jr nz, .done
 
+	; Keep glow state synchronized for newly spawned/reused object structs too.
+if !DEF(MONOCHROME)
```
Rangi42: Please don't gate so much code behind the non-monochrome build. It's simpler to perform the glow/fade logic even in the monochrome build, and gate a minimal change for it, e.g. a dummy-out early `ret` in whichever function actually *applies* the change. That way we don't accidentally introduce significant performance differences or side-effect differences in the monochrome build.

engine/gfx/dynamic_pals.asm
```diff
@@ -97,7 +113,38 @@ ScanObjectStructPals:
...
 	ld a, [hl]
 	and a
-	jr z, .skip
+	jmp z, .skip
+
+	; Look up this object's collision-driven glow type.
+if !DEF(MONOCHROME)
+	ld a, b
+	cpl
+	add NUM_OBJECT_STRUCTS + 1
```
Rangi42: This optimal pattern is not quite immediately obvious, so comments are nice, as we have elsewhere.
```suggestion
	add NUM_OBJECT_STRUCTS + 1 ; a = NUM_OBJECT_STRUCTS - a
```

engine/gfx/dynamic_pals.asm
```diff
@@ -97,7 +113,38 @@ ScanObjectStructPals:
...
+	push de
+	ld e, a
+	ld d, 0
+	ld hl, wObjectGlowTypes
+	add hl, de
```
Rangi42: Save a cycle (https://github.com/pret/pokecrystal/wiki/Optimizing-assembly-code#set-a-16-bit-register-to-a-plus-a-constant):
```suggestion
	add LOW(wObjectGlowTypes)
	ld l, a
	adc HIGH(wObjectGlowTypes)
	sub l
	ld h, a
```

engine/gfx/dynamic_pals.asm
```diff
@@ -223,6 +270,27 @@ MarkUsedPal:
...
+	push hl
+	push bc
+	ld b, 0
+	ld hl, wLoadedObjPalPrevGlows
+	add hl, bc
+	ld a, [wPrevNeededObjPalGlow]
+	cp [hl]
+	pop bc
+	pop hl
+	jr nz, .type_mismatch
```
Rangi42: You should be able to add this:
```diff
+.glow_mismatch
+	pop bc
+	pop hl
 .type_mismatch
```
And then do this, to avoid pushing back what you just popped:
```suggestion
	; The same base palette may need separate glow-transition slots.
	push hl
	push bc
	ld b, 0
	ld hl, wLoadedObjPalGlows
	add hl, bc
	ld a, [wNeededObjPalGlow]
	cp [hl]
	jr nz, .glow_mismatch
	ld hl, wLoadedObjPalPrevGlows
	add hl, bc
	ld a, [wPrevNeededObjPalGlow]
	cp [hl]
	jr nz, .glow_mismatch
	pop bc
	pop hl
```
I also removed a now-redundant `ld b, 0` in there.

engine/gfx/dynamic_pals.asm
```diff
@@ -370,4 +448,154 @@ CheckDualObjectPals:
...
+
+.done
+	pop af
+	ldh [rWBK], a
+	pop hl
+	pop de
+	pop bc
+	pop af
+endc
+	ret
```
Rangi42: Get rid of the monochrome build gating, and you can just `jmp PopAFBCDEHL`. (Although you'll also need to rearrange the initial pushes in the conventional order.)

engine/gfx/dynamic_pals.asm
```diff
@@ -370,4 +448,154 @@ CheckDualObjectPals:
...
+
+if !DEF(MONOCHROME)
```
Rangi42: Honestly, just stop even trying to cater to the monochrome build. We haven't been regularly testing it, we don't publish it, and I'm sure we have some overworld palettes which don't appropriately account for it. If/when I want to test and clean up the monochrome build, I'll deal with palette fading+glowing for it then.

engine/gfx/dynamic_pals.asm
```diff
@@ -370,4 +448,154 @@ CheckDualObjectPals:
...
+	add hl, de
+	ld a, [hl]
+.got_collision
+	pop hl
+	cp COLL_CAMPFIRE_GLOW
+	jr z, .campfire
+	cp COLL_AQUARIUM_GLOW
+	ld a, OBJ_GLOW_AQUARIUM
+	jr z, .got_glow
+	ld a, OBJ_GLOW_NONE
```
Rangi42:
```suggestion
	xor a ; OBJ_GLOW_NONE
```

engine/gfx/sprite_palettes.asm
```diff
@@ -96,6 +96,29 @@ CopySpritePalToOBPal7:
...
+	xor a
+	ld [wNeededObjPalGlow], a
+	ld [wPrevNeededObjPalGlow], a
+	push hl
+	push bc
+	ld a, e
+	sub LOW(wOBPals1)
+rept 3
+	srl a
+endr
```
Rangi42: Use the optimal pattern (https://github.com/pret/pokecrystal/wiki/Optimizing-assembly-code#divide-a-by-8-shift-a-right-3-bits):
```suggestion
	rrca
	rrca
	rrca
	and %00011111
```

engine/gfx/sprite_palettes.asm
```diff
@@ -435,6 +468,144 @@ CalculateStates:
...
+	jr nz, .color_loop
+	pop af
+	ldh [rWBK], a
+.done
+	pop hl
+	pop de
+	pop bc
+	pop af
+endc
+	ret
```
Rangi42, line 503: Again, use `PopAFBCDEHL`.

engine/gfx/sprite_palettes.asm
```diff
@@ -435,6 +468,144 @@ CalculateStates:
...
+	push de
+	push af
+	ld a, c
+	add a
+	add c
+	add b
+	ld e, a
+	ld d, 0
+	ld hl, .adjustments
+	add hl, de
```
Rangi42: Use the optimal pattern (https://github.com/pret/pokecrystal/wiki/Optimizing-assembly-code#set-a-16-bit-register-to-a-plus-a-constant) and you can remove the `push/pop` for `de` too:
```suggestion
	add LOW(.adjustments)
	ld l, a
	adc HIGH(.adjustments)
	sub l
	ld h, a
```

Rangi42 (changes requested):

engine/gfx/dynamic_pals.asm
vulcandth: Applied in d6d6cfa725: current and previous glow comparisons now share one saved HL/BC pair and branch through .glow_mismatch, with the redundant ld b, 0 removed.

engine/gfx/dynamic_pals.asm
vulcandth: Applied in d6d6cfa725: added the requested comment documenting the complemented-counter object index calculation.

engine/gfx/dynamic_pals.asm
vulcandth: Applied in d6d6cfa725, following the later clarification as well: all newly introduced monochrome gates were removed, so collision tracking, palette-slot selection, and fade lifecycle use the same path.

engine/gfx/dynamic_pals.asm
vulcandth: Applied in d6d6cfa725: removed the gate, reordered saves to the conventional HL/DE/BC/AF order, and return through PopAFBCDEHL.

engine/gfx/dynamic_pals.asm
vulcandth: Applied in d6d6cfa725: removed all monochrome-specific gating introduced by this PR; there is no special no-op path.

engine/gfx/sprite_palettes.asm
vulcandth: Applied in d6d6cfa725: the adjustment lookup uses the A-plus-constant pattern and no longer saves/restores DE.

engine/gfx/dynamic_pals.asm
vulcandth: Applied in d6d6cfa725: the no-glow fallback now uses xor a with the OBJ_GLOW_NONE comment.

engine/gfx/sprite_palettes.asm
vulcandth: Applied in d6d6cfa725: replaced the three shifts with the rrca/rrca/rrca plus mask pattern.

constants/gfx_constants.asm
vulcandth: Applied in d6d6cfa725: the glow IDs now use const_def/const, with NUM_OBJ_GLOW_TYPES derived from const_value.

engine/gfx/dynamic_pals.asm
vulcandth: Applied in d6d6cfa725: wObjectGlowTypes is now addressed with the optimal A-plus-constant pattern.

engine/gfx/sprite_palettes.asm
vulcandth: Applied in d6d6cfa725: ApplyObjectGlowToPalette now saves registers conventionally and returns through PopAFBCDEHL.

engine/gfx/dynamic_pals.asm
```diff
@@ -8,6 +8,14 @@ ClearSavedObjPals::
 	ld [wUsedObjectPals], a
 	ld [wNeededPalType], a
 	ld [wLoadedObjPalType], a
+	ld [wNeededObjPalGlow], a
+	ld [wPrevNeededObjPalGlow], a
+	ld hl, wLoadedObjPalGlows
+	ld bc, 16
+	rst ByteFill
+	ld hl, wObjectPrevGlowTypes
+	ld bc, NUM_OBJECT_STRUCTS + 1 ; also clear wObjectGlowFadeActive
```
Rangi42, line 18:
```suggestion
	assert wObjectPrevGlowTypes + NUM_OBJECT_STRUCTS == wObjectGlowFadeActive
	ld bc, NUM_OBJECT_STRUCTS + 1 ; also clear wObjectGlowFadeActive
```

engine/gfx/dynamic_pals.asm
```diff
@@ -370,4 +439,146 @@ CheckDualObjectPals:
...
+	jr z, .next
+	ld a, [wNeededObjPalGlow]
+	and a
+	ld a, [hl]
+	jr z, .mark_transition
+	xor a
+	jr .store_previous
+.mark_transition
+	or OBJ_GLOW_TRANSITION
+.store_previous
```
Rangi42, line 551: Saves a byte and is IMO clearer.
```suggestion
	and a
	ld a, 0 ; no-optimize a = 0
	jr nz, .store_previous
	ld a, [hl]
	or OBJ_GLOW_TRANSITION
.store_previous
```

engine/gfx/sprite_palettes.asm
```diff
@@ -435,6 +469,135 @@ CalculateStates:
...
+	ld a, [wNeededObjPalGlow]
+	ret
+
+ApplyObjectGlowToPalette:
+; Add cool aquarium light to one dynamically loaded object palette. Campfire
+; glow is handled by selecting the daytime source palette in LookupOBPalette.
+	push hl
+	push de
+	push bc
+	push af
```
Rangi42: The sole call site of `ApplyObjectGlowToPalette` looks like all it needs to preserve is `hl`, not the other three registers.

(I see a lot of push/pop operations in this PR, and am slightly concerned about performance.)

Rangi42 (changes requested):

Rangi42:
Confirmed that the glow effect applies to standing NPCs:

image

Rangi42 (commented):
My basic testing of this, around the campfire and the aquarium, works so far. Nothing looks drastically wrong with the code. I would still like Codex to try and do fewer push/pop operations -- CPU usage from just walking around in the overworld is pretty high.

https://github.com/user-attachments/assets/f717c503-6744-45a8-829d-0d69962283bc

Rangi42:
Okay, the CPU graph looks pretty similar in `master` too, without this PR feature. Then again, we already have some amount of maybe-inefficient overcast/dynamic-pal/pal-fade/pal-swap code going on in the overworld loop, and I think every bit of optimization helps here. Minimizing push/pop stack ops is also worthwhile since we can look at a register being pushed/popped and be confident that there's a purpose to it.

vulcandth:
Completed the final optimization review in `8b58693e38`. I reviewed the full discussion and inline feedback, then audited the glow path by register liveness rather than removing stack operations mechanically.

The main changes are:

- `UpdateObjectGlowPals` no longer saves/restores `af`, `bc`, `de`, and `hl` every frame; its only caller does not consume those register values afterward.
- `RefreshObjectGlowTypes` advances the object-struct pointer directly, avoiding a `push hl` / `pop hl` pair per object.
- `ScanObjectStructPals` addresses the current and previous glow arrays directly instead of temporarily saving `de`.
- `MarkUsedPal` now keeps the requested palette ID and shift counters in free registers and directly addresses per-slot metadata. This removes the nested saves during palette/type/glow matching and slot marking. The one `bc` save around `CopyObjectSpritePalHandler` remains because that call path can clobber `bc`, while the selected slot in `c` is still live afterward.
- The fixed aquarium transform is unrolled for the three affected colors and operates directly on the BGR555 channels. This removes the per-color loop save, the helper/table lookup, and its nested saves. Color 3 remains untouched.
- The remaining glow-path stack operations have specific purposes: restoring `rWBK`, preserving the active glow-array pointer across collision lookup, or retaining live values across a clobbering call.

I also checked the edited files for `jp`; none were added (and the new control flow uses `jr`/`jmp` as appropriate).

Validation:

- `python3 utils/optimize.py`: `Found 0 instances.`
- Normal build: passes, with 34,472 bytes free.
- Aquarium transform: exhaustively compared against the intended `R+2, G+5, B+9` saturating BGR555 formula for all 32,768 input colors; all matched.
- PyBoy smoke test: 600 boot frames completed.
- `git diff --check`: passes.

Thanks for the testing and the optimization direction.

vulcandth:
Added the requested map-connection palette fade override in `e1778db377`.

The new `no_connection_palette_fade MAP_A, MAP_B` data entry applies automatically in both directions. `RUGGED_ROAD_NORTH` / `RUGGED_ROAD_SOUTH` is the initial pair.

When an overridden connection is crossed, the connection setup still calculates and stores the destination overcast state, but clears any fade already in progress and immediately applies the destination palettes. Clearing the existing fade state is important for quickly recrossing the boundary; merely declining to initialize another fade would allow the prior transition to continue.

Other map connections retain their current palette-fade behavior, and non-connection fades are not globally disabled.

Validation:

- `python3 utils/optimize.py`: `Found 0 instances.`
- Normal build passes with 34,149 bytes free.
- Confirmed the generated table contains both Rugged Road directions.
- PyBoy completed a 600-frame boot smoke test.
- `git diff --check` passes, and no `jp` instruction was added.

Rangi42 (approved):

Rangi42:
Thanks a lot! There will be more places to take advantage of this feature. :)

PR #1617, opened 2026-09-02
Clean up BG map-related functions

ISSOtm:
**This PR should be reviewed commit by commit**, as it performs both “simple” cleanups *and* more involved functionality changes, and they stack on top of each other. (I think it's also a better idea not to squash it, too.)

I haven't tested this *too* thoroughly, but I've confirmed that I can go through the title sequence and then into all of the START Menu's screens without any issues, which seems like it exercises all code touched by this PR.

This saves 169 bytes of ROM0 (from 77 to 246) without any changes to the rest of the codebase, while also making this code massively simpler and more maintainable.

### DIFF
--- a/constants/ram_constants.asm
+++ b/constants/ram_constants.asm
@@ -547,10 +547,8 @@ DEF LCDC_DEFAULT EQU LCDC_ON | LCDC_WIN_9C00 | LCDC_WIN_ON | LCDC_BLOCK21 | LCDC
 
 ; hBGMapMode::
 	const_def
-	const NO_BG_MAP_TRANSFER   ; 0
-	const TRANSFER_TILEMAP     ; 1: from wTilemap to 0:[hBGMapAddress]
-	const TRANSFER_ATTRMAP     ; 2: from wAttrmap to 1:[hBGMapAddress]
-	const TRANSFER_TILEMAP1    ; 3: from wTilemap to vBGMap1
-	const TRANSFER_ATTRMAP1    ; 4: from wAttrmap to vBGMap3
-	const TRANSFER_TILEMAP_OFS ; 5: like TRANSFER_TILEMAP, but starts on the [hBGMapHalf]th row
-	const TRANSFER_ATTRMAP_OFS ; 6: like TRANSFER_ATTRMAP, but starts on the [hBGMapHalf]th row
+	const NO_BG_MAP_TRANSFER ; 0
+	const TRANSFER_TILEMAP   ; 1: from wTilemap to 0:[hBGMapAddress]
+	const TRANSFER_ATTRMAP   ; 2: from wAttrmap to 1:[hBGMapAddress]
+	const TRANSFER_TILEMAP1  ; 3: from wTilemap to vBGMap1
+	const TRANSFER_ATTRMAP1  ; 4: from wAttrmap to vBGMap3
--- a/engine/gfx/copy_tilemap_at_once.asm
+++ b/engine/gfx/copy_tilemap_at_once.asm
@@ -58,33 +58,6 @@ _SafeCopyTilemapAtOnce::
 	ldh [hMapAnims], a
 	ret
 
-_CopyTilemapAtOnce::
-	ldh a, [hBGMapMode]
-	push af
-	ldh a, [hMapAnims]
-	push af
-
-	xor a
-	assert NO_BG_MAP_TRANSFER == 0
-	ldh [hBGMapMode], a
-	ldh [hMapAnims], a
-
-	di
-	hlcoord 0, 0, wAttrmap
-	ld a, 1 ; BANK(vStandingFrameTiles)
-	call CopyFullTilemapInHBlank
-	hlcoord 0, 0
-	xor a ; BANK(vObjTiles)
-	call CopyFullTilemapInHBlank
-
-	ei ; in case we've passed vblank
-
-	pop af
-	ldh [hMapAnims], a
-	pop af
-	ldh [hBGMapMode], a
-	ret
-
 VBlankSafeCopyTilemapAtOnce::
 	ldh a, [hSCX]
 	ldh [rSCX], a
@@ -266,4 +239,31 @@ CopyTilemapInHBlank:
 	ld sp, wSPBuffer
 	pop hl
 	ld sp, hl
+	ret
+
+_CopyTilemapAtOnce::
+	ldh a, [hBGMapMode]
+	push af
+	ldh a, [hMapAnims]
+	push af
+
+	xor a
+	assert NO_BG_MAP_TRANSFER == 0
+	ldh [hBGMapMode], a
+	ldh [hMapAnims], a
+
+	di
+	hlcoord 0, 0, wAttrmap
+	ld a, 1 ; BANK(vStandingFrameTiles)
+	call CopyFullTilemapInHBlank
+	hlcoord 0, 0
+	xor a ; BANK(vObjTiles)
+	call CopyFullTilemapInHBlank
+
+	ei ; in case we've passed vblank
+
+	pop af
+	ldh [hMapAnims], a
+	pop af
+	ldh [hBGMapMode], a
 	ret
--- a/engine/gfx/copy_tilemap_at_once.asm
+++ b/engine/gfx/copy_tilemap_at_once.asm
@@ -48,9 +48,14 @@ _SafeCopyTilemapAtOnce::
 	ldh [hOAMUpdate], a
 .noForceOAMUpdate
 	bit 3, b
-	ld a, 3
+def NB_ROWS_TILEMAP_ONLY equ 9
+; If we're copying both maps, we won't have enough time to copy as many rows.
+; Thus, the remainder gets copied outside of VBlank, in chunks of 5 rows; this leaves a remainder of 3,
+; which is suitable to be transferred during VBlank (twice: once for each of the two maps).
+def NB_ROWS_BOTH_MAPS equ 3
+	ld a, NB_ROWS_BOTH_MAPS
 	jr z, .gotRowCount
-	ld a, 9
+	ld a, NB_ROWS_TILEMAP_ONLY
 .gotRowCount
 	ldh [hBGMapCopyNRows], a
 	ld a, b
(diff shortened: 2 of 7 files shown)


constants/ram_constants.asm
```diff
@@ -544,3 +544,11 @@ DEF NUM_SRAM_BANKS EQU 4
...
 ; rLCDC::
 DEF LCDC_DEFAULT EQU LCDC_ON | LCDC_WIN_9C00 | LCDC_WIN_ON | LCDC_BLOCK21 | LCDC_BG_9800 | LCDC_OBJ_8 | LCDC_OBJ_ON | LCDC_PRIO_ON
+
+; hBGMapMode::
+	const_def
+	const NO_BG_MAP_TRANSFER
+	const TRANSFER_TILEMAP ; From wTilemap to 0:[hBGMapAddress].
+	const TRANSFER_ATTRMAP ; From wAttrmap to 1:[hBGMapAddress].
+	const TRANSFER_TILEMAP1 ; From wTilemap to vBGMap1.
+	const TRANSFER_ATTRMAP1 ; From wAttrmap to vBGMap3. (TODO: bad name!)
```
Rangi42: Adding a "`0`" suffix makes them obviously pair with the "`1`"-suffixed ones, and goes with hardware.inc's `TILEMAP0/1` and `ATTRMAP0/1` constants. Also, more nitpicky, but I simplified the comments to our more typical lightweight style for same-line trailing comments (not full sentences).
```suggestion
	const TRANSFER_TILEMAP0 ; from wTilemap to 0:[hBGMapAddress]
	const TRANSFER_ATTRMAP0 ; from wAttrmap to 1:[hBGMapAddress]
	const TRANSFER_TILEMAP1 ; from wTilemap to vBGMap1
	const TRANSFER_ATTRMAP1 ; from wAttrmap to vBGMap3
```

ram/hram.asm
```diff
@@ -114,19 +114,12 @@ hSCY:: db
...
-	db
-hBGMapMode::
-; 0 - top third
-; 1 - middle third
-; 2 - bottom third
-	db
-hBGMapHalf::     db
+hTilemapAtomicCopyFlags:: ; Copy of the flags passed to 
+hNbRowsToCopy:: ; Temporary counter for `CopyTilemapInHBlank`.
+hNbTilesToCopy:: db ; Temporary counter for the `Request*bpp` functions.
```
Rangi42: "`Nb`" for "Number" is a very ISSOtm-specific naming convention that pret and PC don't otherwise use ("`Num`" is typical)... but whatever, we have a decade of multiple devs' not-perfectly-consistent naming conventions already. :P

home/gfx2.asm
Rangi42, line 1: Just FYI, these were separated because they didn't used to be contiguous. I believe Prism had placed its gfx2 code in the [often-overlooked](https://github.com/pret/pokecrystal/wiki/Tips-and-tricks#use-more-space-in-the-home-bank) space between the interrupts and the cartridge header.

engine/gfx/copy_tilemap_at_once.asm
```diff
@@ -107,44 +68,53 @@ VBlankSafeCopyTilemapAtOnce::
...
+	ld hl, wAttrmap
+	call CopyTop3MapRows
+	; The previous function returns with VRA0 loaded.
+	ld hl, wTilemap
+	call CopyTop3MapRows
+
+	call PushOAM ; This is still a VBlank handler, after all :)
+
+	; Perform the rest of the copies in HBlank.
+FOR row_idx, 3 /* rows already copied during VBlank */, SCREEN_HEIGHT, 5
```
Rangi42, line 108: Nitpick: we only capitalize `MACRO/ENDM` and sometimes `DEF/...`, not `if/endc/rept/for/endr`.

Less nitpicky: please don't use `/* block comments */`. If they were multi-line they'd potentially break asm-reading scripts like optimize.py and farcheck.py; and when they're in the middle of lines like this one, they break this human's ease of reading code. :P

home/video.asm
```diff
@@ -149,127 +150,39 @@ UpdateBGMap::
...
-.AttributeMapTop
-	coord sp, 0, 0, wAttrmap
-; Next time: bottom half
-	jr .AttributeMapTopContinue
+	xor 1
+	jr nz, :+
+	; Bottom half!
+	add sp, 127 :: add sp, HALF_HEIGHT * SCREEN_WIDTH - 127 ; Source.
+	ld de, HALF_HEIGHT * TILEMAP_WIDTH :: add hl, de ; Destination.
+:
```
Rangi42, line 180: Please do *not* use anonymous labels or `::`! Anonymous labels don't go in the .sym file, so the debugger shows that `jr` as going to some numeric offset. And `::` breaks optimize.py scanning.

Plus both features can be more difficult to read: it's easier to see `jr nz, .skip` and automatically notice the corresponding `.skip`, and for people used to mentally parsing asm one line at a time, `::` breaks that assumption. (I can usually trust that even a long line is only one instruction, possibly with a long formula for its operand; but if a long line has at least one `::` in it, now the whole thing has to be read to see where the instruction boundaries are.)

Rangi42 (changes requested):
Some bugs from testing this:

- [ ] Bill's PC storage is corrupt:
  Image
- [ ] Moving the cursor from the Pokégear's radio to the phone sometimes leaves the textbox looking yellow.
  Image

constants/ram_constants.asm
ISSOtm: True, though nothing requires `hBGMapAddress` to point to TILEMAP0, so I didn't want to name it that way.

ISSOtm: `vBGMap3` is a leftover comment, as I intend to rename that to `vAttrmap1`. I should add that commit to this PR, anyway.

ram/hram.asm
ISSOtm: I don't mind aligning that convention!

home/gfx2.asm
ISSOtm: The fact that those are managed manually is horrifying. They should be split into separate sections, and RGBLINK allowed to pick whatever fits best there. (But you've already said you don't want a “sectionful refactor”, despite all its benefits, so I won't push.)

engine/gfx/copy_tilemap_at_once.asm
ISSOtm: The goal is to explain what the 3 is, since using a constant would not help (shall I write `call Copy{NUM_VBLANK_ROWS}Rows`? :P) and placing the comment elsewhere makes it easier to forget to update it in the event that the code is later changed.

It does make the code harder for you to read right now... but how much of that is a mere matter of habit?

home/video.asm
ISSOtm: As you wish, but I used anon labels because naming is hard and adding numeric suffixes is equally as distracting. (There's a function I've edited somewhere that has some `.nowait3` label, and no `.nowait`, `.nowait1`, or `.nowait2`. That's confusing.)

## Instruction joining

`::` not being supported by `optimize.py` is IMO a limitation that it ought to overcome :P (If what it takes to get this patch accepted is to add support for them, then I'll do it!)

As for the “one line at a time” assumption, I think that's solved through syntax highlighting (the instruction name displays quite differently from the registers!), and that this is actually *upheld* in that both of those instructions form a single `add sp, HALF_HEIGHT * SCREEN_WIDTH` *logical* line, even if it's composed of two *instructions*. Like I've mentioned in https://github.com/Rangi42/polishedcrystal/pull/1617#discussion_r3926586573, my own experience is that it is largely a matter of habit.

If I write the instructions all inline, the grouping is lost and the logic is harder to make out:

```asm
    jr nz, .topHalf
    ; Bottom half!
    add sp, 127
    add sp, HALF_HEIGHT * SCREEN_WIDTH - 127
    ld de, HALF_HEIGHT * TILEMAP_WIDTH
    add hl, de
.topHalf
```

If I try re-introducing the comments, then the first one incorrectly melds with the block's:

```asm
    jr nz, .topHalf
    ; Bottom half!
    ; Source.
    add sp, 127
    add sp, HALF_HEIGHT * SCREEN_WIDTH - 127
    ; Destination.
    ld de, HALF_HEIGHT * TILEMAP_WIDTH
    add hl, de
.topHalf
```

If I try solving this with blank lines, then the final label is [an orphan](https://en.wikipedia.org/wiki/Widows_and_orphans) *and* the entire short block takes up more more unnecessary room:

```asm
    jr nz, .topHalf
    ; Bottom half!

    ; Source.
    add sp, 127
    add sp, HALF_HEIGHT * SCREEN_WIDTH - 127

    ; Destination.
    ld de, HALF_HEIGHT * TILEMAP_WIDTH
    add hl, de

.topHalf ; Blank line? No blank line?
```

Thus, since the lines remain short *and* cohesive, I have opted for joining related instructions on the same lines as the most practical solution.

constants/ram_constants.asm
Rangi42: Please don't -- the value-add of different names for those generic VRAM labels is minimal, and it's more useful to have familiar names that match other pokecrystal-based projects.

engine/gfx/copy_tilemap_at_once.asm
Rangi42: `for row_idx, 3, SCREEN_HEIGHT, 5 ; skip the 3 rows already copied during VBlank` would be fine.

Also just use `x`, we commonly use `x` and `n` as temporary `for` variables.

home/video.asm
Rangi42: Not every dev who works on polishedcrystal (or its forks) is a fan of `::` grouping, or familiar with it at all. I'd like to stick with our conventions of "no block comments, anonymous labels, same-line instructions, or redundant `a` arguments".

constants/ram_constants.asm
ISSOtm: The value-add is large, because those *aren't background maps*, they are attribute maps. The difference is very significant, because otherwise the BG map code only doing `ld sp, vBGMap1` for *both* tilemap and attrmap copy is nonsensical—the only reason it works is that `vBGMap3` is at the same address, but in another bank, and nothing communicates that.

I'm speaking from experience here, this made me initially very confused as to how this code would work, and I am pretty sure I am not alone. (I will thus defend this change in upstream also.)

engine/gfx/copy_tilemap_at_once.asm
ISSOtm: I believe it's worse due to repeating the `3` and thus being far more prone to being missed in any modification, but I don't care about it strongly enough so let's go with that.

As for the variable name, this is as much of an anti-pattern as abusing `c`, `m`, `n`, `i`, `j`, `k`, `l` and so on as loop names in any higher-level language (chiefly C and Python, AFAIK). If the variable can have a name that expresses its semantics well, why shouldn't it? This makes the entire loop body easier to read.

home/video.asm
ISSOtm: Fine, but then please pick which one of the three above alternatives you want.

ram/hram.asm
ISSOtm: Ah, wait, you have already cherry-picked that commit as-is. Oh well...

home/video.asm
Rangi42: How about:
```asm
	jr nz, .topHalf
	; bottom half source
	add sp, 127
	add sp, HALF_HEIGHT * SCREEN_WIDTH - 127
	; bottom half destination
	ld de, HALF_HEIGHT * TILEMAP_WIDTH
	add hl, de
.topHalf
```

PR #1618, opened 2026-09-03
[Needs optim] Have each ROM bank report its own bank number

ISSOtm:
The hope was that this would reduce the number of bankswitch instructions throughout the ROM, but it turns out that there are more *reads* (each one byte larger) than *writes* (two bytes shorter each).

Before:
```
ROM0: 16307 bytes used / 77 free
ROMX: 2049659 bytes used / 14725 free in 126 banks
```
After:
```
ROM0: 16352 bytes used / 32 free
ROMX: 2049791 bytes used / 14593 free in 126 banks
```

Thus, I'm expecting this PR to be rejected. Nonetheless, I want all the work and results to be available *somewhere*, so they can be referenced later should anyone else get a similar idea.

### DIFF
--- a/engine/battle/core.asm
+++ b/engine/battle/core.asm
@@ -8770,7 +8770,7 @@ CopyBackpic:
 	ldh [rWBK], a
 	ld hl, vTiles0
 	ld de, vTiles2 tile $31
-	ldh a, [hROMBank]
+	ld a, [CurROMBank]
 	ld b, a
 	ld c, 7 * 7
 	call Get2bpp
--- a/engine/pc/bills_pc_ui.asm
+++ b/engine/pc/bills_pc_ui.asm
@@ -473,7 +473,7 @@ _BillsPC_SetCursorMode:
 INCLUDE "gfx/pc/cursor.pal"
 
 BillsPC_SafeRequest1bppInWRA6::
-	ldh a, [hROMBank]
+	ld a, [CurROMBank]
 	ld b, a
 	call RunFunctionInWRA6
 .Function:
@@ -484,7 +484,7 @@ BillsPC_SafeRequest1bppInWRA6::
 	jr .Function
 
 BillsPC_SafeRequest2bppInWRA6::
-	ldh a, [hROMBank]
+	ld a, [CurROMBank]
 	ld b, a
 	call RunFunctionInWRA6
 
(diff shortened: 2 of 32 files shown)


Rangi42:
Thanks, I've cherry-picked the unrelated changes from this, and can leave it around as a draft for reference.

ISSOtm:
Consider giving it the `help wanted` tag, so it's clearer that the original author isn't the only one expected to iterate on it?

Rangi42:
What help would be wanted? Isn't this already a complete and coherent PR, just not (yet) improving size or cycle metrics?

ISSOtm:
Yes, thus, it would be help in figuring out a way to make this change improve at least one metric. It's less “help” and more “iteration”/“improvement”, but the label is `help wanted` so that's close enough (it's common signaling for “hey, this PR can use anyone's eyes”).

engine/battle/core.asm
```diff
@@ -8770,7 +8770,7 @@ CopyBackpic:
 	ldh [rWBK], a
 	ld hl, vTiles0
 	ld de, vTiles2 tile $31
-	ldh a, [hROMBank]
+	ld a, [CurROMBank]
 	ld b, a
 	ld c, 7 * 7
 	call Get2bpp
```
vulcandth: Could possible do this for some of these reads instead.

```suggestion
	lb bc, BANK(@), 7 * 7
	call Get2bpp
```

vulcandth, line 8776: For these current bank `Get2bpp` sites...


```suggestion
	ld c, 7 * 7
	call Get2bppInCurrentBank
```

engine/pc/bills_pc_ui.asm
```diff
@@ -473,7 +473,7 @@ _BillsPC_SetCursorMode:
 INCLUDE "gfx/pc/cursor.pal"
 
 BillsPC_SafeRequest1bppInWRA6::
-	ldh a, [hROMBank]
+	ld a, [CurROMBank]
 	ld b, a
```
vulcandth, line 477:
```suggestion
	ld b, BANK(@)
```

vulcandth (commented):
Then we could add something like...

```asm
Get2bppInCurrentBank::
    ldh a, [rLCDC]
    bit B_LCDC_ENABLE, a
    jr nz, Request2bpp.Function
    jr Copy2bpp.Function
```

*untested/unverified*... you should get the idea...

PR #1622, opened 2026-09-05
Improve overworld performance

vulcandth:
The following work was performed with ChatGPT Astra.

## Overworld performance

Walking previously rebuilt all 720 bytes of the visible tile/attribute buffers on each step, then copied only the entering edge and player patch for VBlank. The new step renderer decodes those 88–96 bytes directly into the existing upload queues. VRAM retains the scrolling screen; full map loading, reanchoring, text/menu restoration and explicit refreshes continue to rebuild the full buffers.

The universal sprite limiter remains active every overworld frame for **all 40 OAM entries**, including dry weather. Its single reverse-OAM admission pass protects the player/NPC entries at the end while retaining the drawing priority of weather particles at the beginning. There is no weather-only gate or reduced particle allowance. Clipped sprites are counted without overrunning scratch memory.

The PR also avoids redundant object/rectangle palette work and skips hatch-ability/item calculations when the party has no eggs. Chained palette swaps, NULL entries, fade catch-up, menu restoration, palette-slot exhaustion, and the existing egg-hatching rules are preserved.

## Review updates

Merged master through `a7aa95462cbccc6b768d017213206206623f9af5` in `918b6d973a`. Follow-up `00064615e3` addresses the six published review comments:

* Centralize the coast-sand guard in `UpdateWalkedTilePointers`.
* Document the flag-preserving zero load with `no-optimize`.
* Use checked `row`/`column` macro arguments.
* Relocate the live anchor, standing parities and step direction into HRAM, removing the per-step snapshot copy. All readers/writers were updated; the same six HRAM bytes now hold five live-state bytes and one scratch byte.
* Preserve master's separate, end-aligned `HRAM tail` and genuinely unused space.
* Inline the three party-advance operations at every `DoEggStep` call site.

`python3 utils/optimize.py` across the repository reports **Found 0 instances.** Documentation, profiling tools and generated artifacts remain local and are not part of this PR.

## Updated measurements

Both the latest master and the reviewed candidate were built separately in debug mode with RGBDS 1.0.0, then booted the same Goldenrod battery-save snapshot in unmodified vibeEmu core `4ea2c53cb9caadf5c56d72bdb1f465b715238799`.

Times below are **emulated dots**, not host timings: 4,194,304 dots/second and 70,224 dots/frame. Busy time includes interrupts and VRAM polling but excludes HALT. CPU-budget exceedances are not a dropped-frame counter.

| Metric | Master `a7aa95462c` | Reviewed candidate |
|---|---:|---:|
| Mean map update, dots | 18,456 | 5,722 |
| Mean map update, ms | 4.40 | 1.36 |
| Mean overworld busy dots | 34,434 | 26,615 |
| Maximum overworld busy dots | 91,892 | 68,720 |
| Iterations over the frame CPU budget | 89 / 1,798 | 0 / 1,799 |
| Egg-free cycle in this save, dots | 11,294 | 196 |
| Worst limiter cost in controlled layouts, dots | 33,904 | 6,608 |

The 2,400-frame replay alternates 80-frame left/right/up/down inputs and a 40-frame idle interval. Whole-iteration statistics use the common window after physical frame 600. Master completes 107 steps and the candidate 106; NPC/collision timing can differ, so these are live workload measurements rather than identical call-for-call comparisons.

Identical-input map fixtures show **65.0% / 65.7% / 69.9% / 69.8%** reductions for down/up/left/right. Each direction is another 60 dots cheaper on average than the initially submitted implementation.

The updated 12,000-frame stress replay completes **639 steps**, with **zero over-budget iterations across 11,399 measured iterations**; maximum busy time is 69,842 dots. These results cover the supplied Goldenrod route, not every map, weather effect or scripted event. Intentional menu waits and collision/idle pauses are not walking-frame failures.

## Validation

* Full debug builds of master and candidate; no assembler/linker warnings.
* [GitHub CI passed](https://github.com/Rangi42/polishedcrystal/actions/runs/33939885894) all eight regular/faithful/debug/Virtual Console build variants.
* 8,400 map fixtures covering data seeds, widths, parities, directions, coast states and VRAM wrapping. Compared transfer data/destinations, anchors and bank restoration, then full-screen rebuilds and connection-buffer operations.
* 4,096 independent sprite-limiter reference cases, including all-40 overlap, top/bottom clipping and a scratch-overrun guard.
* 205 object-palette comparisons from live snapshots, including cold/full slots, disabled updates and no-apply mode.
* 8,586 rectangle-swap comparisons covering boundaries, chained/NULL entries, fades and menu invalidation.
* 10,160 egg-cycle comparisons against the original routine, including all party sizes/egg masks, hatching abilities, Oval Charm, abilities disabled and hatch thresholds.
* Goldenrod walking, menu entry/exit and the extended replay without emulator CPU faults or ROM crash codes.

Further in-game review should include dense weather/player visibility, running/cycling, connected map edges, text/menu restoration, coast-sand tracks and map edits such as Cut or bridges. Boot from a battery save when testing the new ROM, rather than restoring an emulator state from the previous binary.


vulcandth:
Note: I didn't commit the `build/overworld-perf`directory.

engine/overworld/player_step.asm
```diff
@@ -454,14 +364,8 @@ UpdateBGMapColumn::
...
-ReloadWalkedTile:
-; Update tile player is to walk on
-	hlcoord 8, 6
-	ld de, wBGMapBuffer
-	call .CommitTiles
-	hlcoord 8, 6, wAttrmap
-	ld de, wBGMapPalBuffer
-	call .CommitTiles
+UpdateWalkedTilePointers:
+; The streamer has already written the walked patch to the upload buffers.
```
Rangi42, line 367: Since all four calls to `UpdateWalkedTilePointers` are conditional on `CheckPlayerCoastSandColl`, that logic could be inlined here instead of at the four call sites:
```suggestion
UpdateWalkedTilePointers:
	call CheckPlayerCoastSandColl
	ret z

; The streamer has already written the walked patch to the upload buffers.
```

engine/overworld/stream_map_part.asm
```diff
@@ -0,0 +1,217 @@
...
+	ld a, [wOverworldMapAnchor + 1]
+	ldh [hStreamMapAnchor + 1], a
+	ld a, [wMetatileStandingX]
+	ldh [hStreamMapStandingX], a
+	ld a, [wMetatileStandingY]
+	ldh [hStreamMapStandingY], a
+	ld a, [wPlayerStepDirection]
+	ldh [hStreamMapDirection], a
+	call CheckPlayerCoastSandColl
+	ld a, 0
```
Rangi42: This might have been caught if it ran `utils/optimize.py`.
```suggestion
	ld a, 0 ; no-optimize a = 0
```

engine/overworld/stream_map_part.asm
```diff
@@ -0,0 +1,217 @@
...
+	ret
+ENDM
+
+StreamMapRow:
+	call StreamMapSetup
+	stream_map_quarters 0
+
+StreamMapColumn:
+	call StreamMapSetup
+	stream_map_quarters 1
```
Rangi42: Since these are sentinels, not meaningful values, prefer `===`-checked "enum" arguments to raw numbers.
```suggestion
StreamMapRow:
	call StreamMapSetup
	stream_map_quarters row

StreamMapColumn:
	call StreamMapSetup
	stream_map_quarters column
```

engine/overworld/stream_map_part.asm
```diff
@@ -0,0 +1,217 @@
...
+	ld a, [wOverworldMapAnchor]
+	ldh [hStreamMapAnchor], a
+	ld a, [wOverworldMapAnchor + 1]
+	ldh [hStreamMapAnchor + 1], a
+	ld a, [wMetatileStandingX]
+	ldh [hStreamMapStandingX], a
+	ld a, [wMetatileStandingY]
+	ldh [hStreamMapStandingY], a
+	ld a, [wPlayerStepDirection]
+	ldh [hStreamMapDirection], a
```
Rangi42: If these need to be in HRAM for efficiency, should we just relocate the variables from WRAM to HRAM, and avoid this whole copy operation?

ram/hram.asm
```diff
@@ -208,7 +208,14 @@ ENDU
...
 
-	ds 11 ; unused
+; Step renderer inputs must remain accessible while WRAM banks 3/4 are mapped.
+hStreamMapAnchor:: dw
+hStreamMapStandingX:: db
+hStreamMapStandingY:: db
+hStreamMapDirection:: db
+hStreamMapWalkedPatch:: db
+
+	ds 5 ; unused
```
Rangi42: Slight merge conflict with master.

Rangi42 (changes requested):

engine/pokemon/breeding.asm
```diff
@@ -168,6 +168,19 @@ DoEggStep::
...
 	ret z
 
+	; Most parties have no eggs. Avoid loading every member's ability and
+	; looking up the Oval Charm when there is no hatch counter to update.
+	ld e, a
+	ld hl, wPartyMon1IsEgg
+.find_egg
+	bit MON_IS_EGG_F, [hl]
+	jr nz, .has_egg
+	call .NextPartyMon
```
Rangi42: This `call` is 3 bytes. The function body, `ld bc, PARTYMON_STRUCT_LENGTH :: add hl, bc :: dec e`, is 5. It's only called a few times. Since speed is important here, just inline it at all the call sites.

engine/overworld/player_step.asm
vulcandth: Done in 00064615e3. The coast-sand check and early return now live in `UpdateWalkedTilePointers`; all four callers call it directly. The differential map fixtures still pass for both coast-sand cases and ordinary movement, including the existing footprint-preservation behavior.

engine/overworld/stream_map_part.asm
vulcandth: Added the `no-optimize a = 0` annotation with the reason: this load must preserve the Z flag returned by `CheckPlayerCoastSandColl`. I also reviewed the optimization guide and ran `python3 utils/optimize.py` across the repository after the changes: **Found 0 instances.**

engine/overworld/stream_map_part.asm
vulcandth: Done in 00064615e3: the invocations now use `row` and `column`, both branches use `===` string comparisons, and the macro asserts that its argument is one of those two values.

engine/overworld/stream_map_part.asm
vulcandth: Agreed. The copy was needed for WRAM-bank visibility, but relocating the live state removes it altogether. `hOverworldMapAnchor`, `hMetatileStandingX/Y`, and `hPlayerStepDirection` now hold the live values, and all readers/writers use those addresses (with `ldh` for direct accumulator accesses). The separate snapshots and their ten load/store instructions are gone. This uses the same HRAM footprint: five live-state bytes plus the walked-patch scratch byte. The former WRAM slots are left unused so unrelated addresses do not shift. Validation includes full map rebuilding and `BufferScreen`/`SaveScreen`/`LoadConnectionBlockData`, as well as walking and menu restoration.

ram/hram.asm
vulcandth: Merged current master (`a7aa95462c`) in 918b6d973a. The resolution preserves the new `HRAM tail` section and its end alignment, and leaves the spare HRAM genuinely unallocated instead of restoring the `ds` padding. The debug build passes without merge conflicts or linker warnings.

engine/pokemon/breeding.asm
vulcandth: Done in 00064615e3: inlined the three instructions at all three call sites and removed `.NextPartyMon`. The six-member egg-free path drops from 316 to 196 emulated dots. All 10,160 comparisons against the original hatch routine still pass, including party sizes/egg positions, hatching abilities, Oval Charm, and hatch thresholds.

engine/overworld/weather.asm
```diff
@@ -1164,105 +1164,60 @@ IsEvenSpriteIndex:
 	ret
 
 WeatherSpriteLimitCheck:
+; Admit sprites in the same reverse-OAM priority order as the old overflow
+; handler. Each candidate touches at most eight counters, once; if it
+; exceeds the ten-sprite limit, undo its tentative counts and hide it.
+; The old code rescanned all 40 sprites for every overflowing scanline.
```
Rangi42: Comments referring to the "old" state of code are at least as bad as commented-out code, in terms of reader confusion. We have Git history for that.
```suggestion
; Admit sprites in reverse-OAM priority order.
; Each candidate touches at most eight counters, once.
; If it exceeds the ten-sprite limit, undo its tentative counts and hide it.
```

engine/overworld/weather.asm
```diff
@@ -1164,105 +1164,60 @@ IsEvenSpriteIndex:
...
 	ldh a, [rWBK]
 	push af
 	ld a, BANK(wWeatherScratch)
 	ldh [rWBK], a
 
-	; Write (sprite limit + 1) to wWeatherScratch.
-	; Call SpriteLimitExceeded if the sprite overlap count reaches zero.
+	assert LOW(wWeatherScratch) == 0
+	assert LOW(wShadowOAM) == 0
```
Rangi42: Please put these assertions next to where they're actually relevant, with a brief comment explaining how these truths result in correct state for `hl` or whichever registers are depending on it.

engine/overworld/weather.asm
```diff
@@ -1164,105 +1164,60 @@ IsEvenSpriteIndex:
...
-	call z, SpriteLimitExceeded
+	jr z, .reject{d:n}
+if n < TILE_WIDTH
 	inc l
+endc
 endr
-	dec [hl]
-	call z, SpriteLimitExceeded
+	jr .next
+for n, TILE_WIDTH, 1, -1
```
Rangi42, line 1202: Astra came up with this unprompted? Nice use of `n` across multiple rgbasm `for` loops instead of `\@`.

home/menu.asm
```diff
@@ -636,6 +636,9 @@ GetMenuDataPointerTableEntry::
 	ret
 
 ClearWindowData::
+	; Text and menus may overwrite a rectangle's selected map palette.
```
Rangi42, line 639: Ooh, did Astra figure out the root cause / a better fix that my brute-force solution? I'll have to test and make sure that swapped palettes don't "blink" in areas like Goldenrod or Saffron.

ram/hram.asm
```diff
@@ -208,6 +208,14 @@ ENDU
 hUsedWeatherSpriteIndex:: db
 hUsedOAMIndex:: db
 
+; Live overworld state must remain accessible while WRAM banks 3/4 are mapped.
```
Rangi42: Yeah, it's generally true of HRAM and WRAM0 that it must remain live across multiple WRAMX banks. Also referring to the numeric bank IDs is not meaningful, versus saying which relevant variables are contained in both (decompressed metatiles and attributes respectively).
```suggestion
```

ram/hram.asm
```diff
@@ -208,6 +208,14 @@ ENDU
...
+hOverworldMapAnchor:: dw
+hMetatileStandingY:: db
+hMetatileStandingX:: db
+hPlayerStepDirection:: db
+; Step renderer scratch.
+hStreamMapWalkedPatch:: db
```
Rangi42:
```suggestion
hStreamMapWalkedPatch:: db ; Step renderer scratch.
```

ram/wramx.asm
```diff
@@ -473,7 +473,7 @@ wPlayerStepFlags::
 ; bit 4: In midair
 ; bits 0-3: unused
 	db
-wPlayerStepDirection:: db
+	ds 1 ; unused
```
Rangi42, line 477:
```suggestion

	ds 1 ; unused
```

Rangi42 (changes requested):

engine/overworld/weather.asm
vulcandth: Yes. Yes it did.

Rangi42:
> Further in-game review should include dense weather/player visibility, running/cycling, connected map edges, text/menu restoration, coast-sand tracks and map edits such as Cut or bridges.

I've tested all of these, and it seems *okay*. Haven't found any new bugs/glitches. CPU usage is still high -- try looking around Snowtop Mountain at night when there are the lit-up windows *and* snow, or running around Rugged Road with the sandstorm *and* campfire glow *and* palette switching when you cross the north/south map connection, sometimes BGB's CPU meter remains at 100% (which it does in master too). However, I trust that you measured real numeric improvement, and I'm not seeing *worse* perf than before. Leaving this open for now in case you have further testing or improvements.

engine/overworld/weather.asm
vulcandth: Done in ad1f65e4c5. Each assertion now sits beside the operation that depends on it: the eight-pixel Y-range calculation, replacing `l` with the scratch offset, and stepping backward through OAM using `e` and carry. The nearby comments explain those register/address assumptions.

Also merged upstream master through ff1c42bb1f. The debug build passes, the optimizer reports zero findings for this file, and the ROM is byte-identical before and after the assertion/comment change on the merged base.

PR #1624, opened 2026-09-05
Prevent TM and HM reuse from restoring PP

### LINKED ISSUE #1577: Prevent TM reuse from giving unlimited PP
When a move is overwritten by a TM, its current PP should not be increased. Otherwise players can reteach TM moves to restore their PP indefinitely.

(When a move is learned by level-up or evolution, it *should* restore PP, since that's not repeatable and so players won't feel discouraged by new moves they can't fully use. Especially since mid-battle evolution is possible.)

vulcandth:
Fixes #1577.

Keep `LEARNING_TM` set around `LearnMove` for both TMs and HMs, then clear it on success or cancellation. Previously, `AskTeachTMHM` cleared it before move learning began, so replacing Hyper Beam at 4 PP with Curse restored Curse to 10 PP. The replacement now keeps 4 PP, capped at the new move's base PP.

Mask out PP Up bits before comparing the old PP. Following Rangi42's review, HMs use the same PP cap as TMs; field use does not depend on remaining PP. Empty move slots and moves learned outside TM/HM teaching, including level-up and evolution, still receive full PP. Also use the actual `HM01` boundary for the boot-up message.

Validation on the revised branch, including the latest `master` commits:
- Reviewed the [assembly optimization guide](https://github.com/pret/pokecrystal/wiki/Optimizing-assembly-code).
- `python3 utils/optimize.py`: zero instances.
- `make -B -j8` and `make -B -j8 faithful`.
- Temporary RGBDS/PyBoy harness executed the source's PP-selection block for 512 combinations of learning flags, current PP, PP Ups, and base PP; all passed. Six TM/HM flag-lifetime cases passed with `LearnMove` stubbed for success/cancellation, requiring the PP-limiting flag for both TMs and HMs and checking cleanup afterward.

Emulator validation covers the assembly blocks; a full interactive playthrough was not performed.

### DIFF
--- a/engine/items/tmhm2.asm
+++ b/engine/items/tmhm2.asm
@@ -370,8 +370,6 @@ AskTeachTMHM:
 	ld a, [hl]
 	push af
 	res NO_TEXT_SCROLL, [hl]
-	ld hl, wForgettingMove
-	set LEARNING_TM_F, [hl]
 	ld a, [wCurTMHM]
 	ld [wTempTMHM], a
 	farcall GetTMHMMove
@@ -381,12 +379,9 @@ AskTeachTMHM:
 	call CopyName1
 	ld hl, Text_BootedTM ; Booted up a TM
 	ld a, [wCurTMHM]
-	cp HM01 + 1 ; off by one error?
+	cp HM01
 	jr c, .TM
 
-	; allow full PP restore for HMs
-	ld hl, wForgettingMove
-	res LEARNING_TM_F, [hl]
 	ld hl, Text_BootedHM ; Booted up an HM
 .TM:
 	call PrintText
@@ -399,8 +394,6 @@ AskTeachTMHM:
 	pop bc
 	ld a, b
 	ld [wOptions1], a
-	ld hl, wForgettingMove
-	res LEARNING_TM_F, [hl]
 	ret
 
 ChooseMonToLearnTMHM:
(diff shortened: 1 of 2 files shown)


vulcandth:
Hmm.... I wonder if some of the HM stuff was needed to prevent softlocks? Some of this may need to be reverted?
@FredrIQ can you provide some insight?

Rangi42:
You don't need HM moves to have PP to use them in the field, and more recently you don't even need to know them at all, just have a mon capable of using them. So this shouldn't affect softlockability.

engine/items/tmhm2.asm
```diff
@@ -484,7 +477,17 @@ TeachTMHM:
 	call KnowsMove
 	jr c, .nope
 
+; Keep the TM flag set while learning, but allow full PP for HMs.
```
Rangi42: No need to allow full PP for HMs, since it doesn't affect their usability on the field.

engine/items/tmhm2.asm
```diff
@@ -381,12 +379,9 @@ AskTeachTMHM:
 	call CopyName1
 	ld hl, Text_BootedTM ; Booted up a TM
 	ld a, [wCurTMHM]
-	cp HM01 + 1 ; off by one error?
+	cp HM01
```
Rangi42: Haha, how long was that "off by one error?" there. :P

Rangi42 (changes requested):

engine/items/tmhm2.asm
vulcandth: Removed the HM exception in e514573d8e. Both TMs and HMs now retain the old move’s current PP, capped at the new move’s base PP. The flag is cleared after success or cancellation. I also checked CheckPartyMove: it does not check PP, so this leaves field usability unchanged. The updated assembly checks cover both TM and HM paths.

engine/items/tmhm2.asm
vulcandth: Since June 26, 2017! Git blame points to 3656d5604c588cb08f36676213a1b1d7da1c4418 ("TM70 Flash says Booted up a TM, not an HM"). The HM01 boundary correction is retained here.

Rangi42: This was not an off-by-one error after all:

Image

Rangi42 (approved):
LGTM -- when you overwrite a move, whether TM or HM, the PP becomes min(current PP, new move's max PP).

PR #1626, opened 2026-09-05
Require explicit full-Bag handling in item-giving macros

### LINKED ISSUE #1526: Update scripting macros for item-giving NPCs to make them fail-safe against a full Bag
Can transfer the items to the PC after wining the battle? Or prevent the battle?

Version: 3.2.3

vulcandth:
Item-giving scripts could omit full-Bag handling without an assembly error. Require a failure command for checked item gifts and make unchecked gifts explicit in the macro name, addressing #1526.

Both `giveitem` and `verbosegiveitem` have four fixed signatures (shown here for the verbose family):

- `verbosegiveitem <item>, <failure command>` gives one item.
- `verbosegiveitems <item>, <quantity>, <failure command>` gives the required quantity.
- `verbosegiveitem_unsafe <item>` gives one item without checking the result.
- `verbosegiveitems_unsafe <item>, <quantity>` gives the required quantity without checking the result.

`verbosegiveitemvar <item>, <variable>, <failure command>` also requires a failure command. RGBDS assertions enforce argument counts and reject empty failure commands. There is no `0` escape hatch, Python verification script, or additional CI step.

Migrate all 226 callers while preserving their quantities, short branches, and text exits. Keep safety comments explaining unchecked refunds, deferred checks, debug inventory, and gifts that allow forfeiture. Marley's existing retry handling on `master` remains intact.

Validation against merged `master` (`882c921394`):

- Reviewed the [assembly optimization guidance](https://github.com/pret/pokecrystal/wiki/Optimizing-assembly-code); `python3 utils/optimize.py` reports **0 instances** before and after the revision.
- `make -j8` and a clean `make -j8 faithful debug vc` pass using RGBDS 1.0.0.
- Standard ROM, Faithful/Debug ROM, Faithful/Debug VC ROM, and VC patch match separate builds of unchanged `master` byte-for-byte.
- Local assembly probes passed 35 bytecode comparisons and rejected 25 malformed calls, including missing arguments and former `0` opt-outs. All 226 migrated calls expand to the original master script commands, including debug-only calls.
- `git diff --check` passes.

Fixes #1526.


.github/workflows/build.yml
```diff
@@ -51,6 +51,9 @@ jobs:
         name: "Checkout Rangi42/polishedcrystal"
         uses: actions/checkout@v7
 
+      - name: "Test item-giving macros"
+        run: python3 utils/test_item_give_macros.py
```
vulcandth: We don't need to verify this in our CI....

Rangi42 (changes requested):
I'd like to avoid overloading one macro with different argument schemes, especially since it then needs a Python script to do the actual verification. How about this set of macros:

- `verbosegiveitem <item>, <failure command>`: assumes quantity 1, failure command is required (no need for that "`if STRCMP("\1", "0")`" unsafe escape hatch)
- `verbosegiveitems <item>, <quantity>, <failure command>`: quantity and failure command are both required (note the pluralized macro name)
- `verbosegiveitem_unsafe <item>`: assumes quantity 1, no failure command, but the "unsafe" cues us to use it safely (e.g. `verbosegiveitem MINT_LEAF, 0` would now be `verbosegiveitem_unsafe MINT_LEAF`, still with the "Commit the challenge result even if the Bag cannot fit the prize" safety comment)
- `verbosegiveitems_unsafe <item>, <quantity>`: quantity is required, no failure command, but the "unsafe" cues us to use it safely (e.g. `verbosegiveitem POKE_BALL, 0, 5` would now be `verbosegiveitems_unsafe POKE_BALL, 5`, still with the "Finish the tutorial even if the Bag cannot fit the bonus balls" safety comment)

That will let us get rid of the Python script.

.github/workflows/build.yml
vulcandth: Removed the added CI step and the Python verification script in 0296413105. The build workflow now matches master.

vulcandth:
@Rangi42 Implemented your proposed split in 0296413105:

- `verbosegiveitem` / `verbosegiveitems` have fixed signatures and require a failure command.
- `verbosegiveitem_unsafe` / `verbosegiveitems_unsafe` explicitly omit the check, retaining the safety comments at their call sites.
- Applied the same split to `giveitem` / `giveitems`. `verbosegiveitemvar` keeps its fixed three-argument signature with a required failure command.
- Removed the `0` escape hatch, handler-dispatch helper, Python verification script, and added CI step. Argument checks are RGBDS assertions.

Merged current `master` (882c921394). Reviewed the optimization guide and reran `utils/optimize.py`: 0 instances. Standard and clean Faithful/Debug/VC builds pass without warnings; the ROMs and VC patch match separate builds of current `master` byte-for-byte. Updated the PR description to reflect the final API and validation.

Rangi42 (approved):
Nice work! The ROM hash is unchanged, and I like this API (after my refactoring of it).

PR #1629, opened 2026-09-05
Fix weather and hidden object palette allocation

### LINKED ISSUE #1621: Ledge-hopping shadow sprite uses the wrong OBJ palette
Looks like the nearby palette glow effect is responsible for this, although it's affecting NPCs, not the player.

https://github.com/user-attachments/assets/94944ed6-2c51-48d4-8625-882a337bbfa1

vulcandth:
Fixes #1621.

On Rugged Road North, a ledge shadow could take OBJ palette 6 while slot 7 was free. Sandstorm uses fixed OBJ 6 and overwrote the shadow's gray palette on the next frame, also briefly rendering sand with the shadow palette. Reserve the weather slot during active weather and particle cooldown, exclude it from object palette matches, and reassign any existing owner before starting weather.

Exclude off-screen and script-hidden objects from palette allocation, refreshing ownership when they leave or return to visibility. The supplied save retained an off-screen fruit tree because of `WONT_DELETE`; its palette is now available to visible objects. Verified that the map's alternate hikers are already excluded by time masks and only one spawns at each time of day.

Keep weather palette metadata consistent when replacing a Pokémon palette. Snow now records its unindexed palette in OBJ 6's metadata before switching WRAM banks, instead of writing OBJ 7's address in palette WRAM.

Validation:
- Reproduced the supplied debug save in vibeEmu before the fix; verified 31 shadow frames remain black while sandstorm retains its palette after the fix.
- Local symbol-driven vibeEmu harness: 36 allocation fixtures covering all weather types, cooldown, existing OBJ 6 matches, repeated scans, and reclaiming the slot after weather. Additional checks cover weather startup, snow metadata, off-screen/reappearing trees, hide/show movement commands, and all four hiker time masks.
- Built standard, faithful, and debug ROMs, plus standard and faithful Virtual Console variants.
- Reviewed the [assembly optimization guide](https://github.com/pret/pokecrystal/wiki/Optimizing-assembly-code). Ran `python3 utils/optimize.py`: 0 instances. `git diff --check` passes.

### DIFF
--- a/engine/events/weather.asm
+++ b/engine/events/weather.asm
@@ -43,6 +43,8 @@ SetCurrentWeather::
 .skip_cooldown
 	ld a, b
 	ld [wCurWeather], a
+	; Move any object out of the fixed weather slot before overwriting it.
+	farcall CheckForUsedObjPals
 	farcall LoadWeatherPal
 	call LoadWeatherGraphics
 	xor a
--- a/engine/gfx/dynamic_pals.asm
+++ b/engine/gfx/dynamic_pals.asm
@@ -89,6 +89,13 @@ CheckForUsedObjPals::
 	ld [wNeededMonPalLight], a
 
 	call CheckDualObjectPals
+	; Weather OAM uses a fixed slot, including particles still clearing after
+	; the weather stops. Do not let an object overwrite that palette.
+	call CheckWeatherPalInUse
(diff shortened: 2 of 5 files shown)


Rangi42:
Tested this and it works! Thanks :)

PR #1630, opened 2026-09-06
Add configurable in-game clock speeds

### LINKED ISSUE #1206: Suggestion to make No RTC's time speed configurable
I would like to play without the real-time clock, but the default rate of 1 in-game minute per 6 seconds is way too long for my liking, and I'd rather not use fast forward or reset the clock. It would be nice to have the options for, say, 1, 2, 4 and 6 seconds per minute.

vulcandth:
Replace “Real-time clock: Yes/No” with “In-game clock: RTC/×6/×12/×24”, following Rangi42’s direction in #1206. The accelerated choices give an in-game day in 4, 2, or 1 hours of play; the play-time counter stays at normal speed.

Encode the four clock choices in bits 3–4 of `wInitialOptions2` (`00` RTC, `01` ×6, `10` ×12, `11` ×24), with evolve-in-battle moved to bit 5. The menu cycles the field directly, and hardware clock access checks the full field. This changes the previous save-option encoding.

Phone and Bug-Catching Contest timers use unaccelerated play time when RTC is disabled, preserving their durations without overflowing their byte-sized minute counters at ×24. The contest now signals timeout at exactly 0:00. `RTC_DAY_CYCLE` names the existing 140-day (20-week) wrap used by clock and elapsed-time routines.

Validation uses the local vibeEmu core to execute assembled ROM routines with injected state and timestamps; it does not automate playing the game. Standard and Faithful builds pass with RGBDS warnings enabled.

| System | Checked on both builds |
| --- | --- |
| Clock rates | Full in-game days take 4/2/1 hours of play at ×6/×12/×24; play time advances normally; paused game logic does not advance the clock. |
| Menu | 768 input/redraw cases verify wrapping, displayed tile characters, and unrelated option bits. |
| Save/reload | All four choices survive the actual option-save and load routines. |
| RTC access | Non-RTC modes preserve hardware time; RTC mode reads/writes it. |
| Phone | All four modes: 20/10/5/3-minute countdowns, clamped delay index, and Lightning Rod doubling (40 scenarios, checked minute by minute). |
| Contest | All four modes: active with 0:01 left at 19:59; timeout carry set at exactly 20:00, across RTC day and play-time hour-byte rollover. |
| Elapsed time | 2,997 second-borrow and 2,997 minute-borrow hour rollovers, plus RTC midnight/day wrap. |
| Daily clock | `FixDays` wraps day 140 to 0; elapsed time from day 139 to 0 is one day. |

Reviewed the [assembly optimization guide](https://github.com/pret/pokecrystal/wiki/Optimizing-assembly-code), including the requested pointer-lookup pattern. `python3 utils/optimize.py` reports zero findings; `git diff --check` passes.

Fixes #1206.

### DIFF
--- a/constants/ram_constants.asm
+++ b/constants/ram_constants.asm
@@ -116,9 +116,9 @@ DEF LINK_OPTMASK EQU (1 << NATURES_OPT) | (1 << ABILITIES_OPT) | (1 << PERFECT_I
 ; wInitialOptions2::
 	const_def 2
 	const NO_EXP_OPT           ; 2
-	const RTC_OPT              ; 3
-	const EVOLVE_IN_BATTLE_OPT ; 4
-	const CLOCK_SPEED_OPT     ; 5; two bits
+	const CLOCK_OPT           ; 3; two bits
+	const_skip
+	const EVOLVE_IN_BATTLE_OPT ; 5
 	const_skip
 	const RESET_INIT_OPTS      ; 7
 
@@ -128,14 +128,12 @@ DEF LINK_OPTMASK EQU (1 << NATURES_OPT) | (1 << ABILITIES_OPT) | (1 << PERFECT_I
 	const EVS_OPT_MODERN   ; %10
 DEF EV_OPTMASK EQU %11
 
-; Keep the existing RTC flag and zero-valued ×6 speed compatible with old saves.
-DEF CLOCK_SPEED_MASK EQU %11 << CLOCK_SPEED_OPT
-DEF CLOCK_OPTMASK EQU (1 << RTC_OPT) | CLOCK_SPEED_MASK
-DEF CLOCK_RTC EQU 1 << RTC_OPT
-DEF CLOCK_6X EQU 0 << CLOCK_SPEED_OPT
-DEF CLOCK_12X EQU 1 << CLOCK_SPEED_OPT
-DEF CLOCK_24X EQU 2 << CLOCK_SPEED_OPT
-DEF NUM_CLOCK_OPTIONS EQU 4
+	const_def 0, 1 << CLOCK_OPT
+	const CLOCK_RTC  ; %00
+	const CLOCK_6X   ; %01
+	const CLOCK_12X  ; %10
+	const CLOCK_24X  ; %11
+DEF CLOCK_OPTMASK EQU %11 << CLOCK_OPT
 
 ; wOptionsMenuDescriptionState::
 	const_def
--- a/constants/ram_constants.asm
+++ b/constants/ram_constants.asm
@@ -118,7 +118,8 @@ DEF LINK_OPTMASK EQU (1 << NATURES_OPT) | (1 << ABILITIES_OPT) | (1 << PERFECT_I
 	const NO_EXP_OPT           ; 2
 	const RTC_OPT              ; 3
 	const EVOLVE_IN_BATTLE_OPT ; 4
-	const_skip 2
+	const CLOCK_SPEED_OPT     ; 5; two bits
+	const_skip
 	const RESET_INIT_OPTS      ; 7
 
 	const_def
@@ -126,6 +127,15 @@ DEF LINK_OPTMASK EQU (1 << NATURES_OPT) | (1 << ABILITIES_OPT) | (1 << PERFECT_I
 	const EVS_OPT_CLASSIC  ; %01
 	const EVS_OPT_MODERN   ; %10
 DEF EV_OPTMASK EQU %11
+
+; Keep the existing RTC flag and zero-valued ×6 speed compatible with old saves.
+DEF CLOCK_SPEED_MASK EQU %11 << CLOCK_SPEED_OPT
+DEF CLOCK_OPTMASK EQU (1 << RTC_OPT) | CLOCK_SPEED_MASK
+DEF CLOCK_RTC EQU 1 << RTC_OPT
+DEF CLOCK_6X EQU 0 << CLOCK_SPEED_OPT
+DEF CLOCK_12X EQU 1 << CLOCK_SPEED_OPT
+DEF CLOCK_24X EQU 2 << CLOCK_SPEED_OPT
+DEF NUM_CLOCK_OPTIONS EQU 4
 
 ; wOptionsMenuDescriptionState::
 	const_def
--- a/constants/ram_constants.asm
+++ b/constants/ram_constants.asm
@@ -116,8 +116,8 @@ DEF LINK_OPTMASK EQU (1 << NATURES_OPT) | (1 << ABILITIES_OPT) | (1 << PERFECT_I
 ; wInitialOptions2::
 	const_def 2
 	const NO_EXP_OPT           ; 2
-	const CLOCK_OPT           ; 3; two bits
-	const_skip
+	const CLOCK_OPT            ; 3 ; two bits
+	const_skip                 ; also used by CLOCK_OPT
 	const EVOLVE_IN_BATTLE_OPT ; 5
 	const_skip
 	const RESET_INIT_OPTS      ; 7
@@ -129,10 +129,10 @@ DEF LINK_OPTMASK EQU (1 << NATURES_OPT) | (1 << ABILITIES_OPT) | (1 << PERFECT_I
 DEF EV_OPTMASK EQU %11
 
 	const_def 0, 1 << CLOCK_OPT
-	const CLOCK_RTC  ; %00
-	const CLOCK_6X   ; %01
-	const CLOCK_12X  ; %10
-	const CLOCK_24X  ; %11
+	const CLOCK_RTC ; %000_00_000
+	const CLOCK_6X  ; %000_01_000
+	const CLOCK_12X ; %000_10_000
+	const CLOCK_24X ; %000_11_000
 DEF CLOCK_OPTMASK EQU %11 << CLOCK_OPT
 
 ; wOptionsMenuDescriptionState::
--- a/engine/menus/initial_options_menu.asm
+++ b/engine/menus/initial_options_menu.asm
@@ -287,24 +287,60 @@ InitialOptions_AffectionBonus:
 	jmp OptionsShared_PlaceStringAtValueCoord
 
 InitialOptions_RTC:
-	ld hl, wInitialOptions2
+	ld a, [wInitialOptions2]
+	and CLOCK_OPTMASK
+	ld hl, .Values
+	ld c, 0
+.find_value
+	cp [hl]
+	jr z, .got_value
+	inc hl
+	inc c
+	bit 2, c
+	jr z, .find_value
+	ld c, 0 ; fall back to RTC for an invalid option value
+.got_value
 	ldh a, [hJoyPressed]
 	and PAD_LEFT | PAD_RIGHT
-	jr nz, .Toggle
-	bit RTC_OPT, [hl]
-	jr z, .SetNo
-	jr .SetYes
-.Toggle:
-	bit RTC_OPT, [hl]
-	jr z, .SetYes
-.SetNo:
-	res RTC_OPT, [hl]
-	ld de, NoString
+	jr z, .input_done
+	bit B_PAD_LEFT, a
+	jr nz, .left
+	inc c
+	jr .wrap
+.left
+	dec c
+.wrap
+	ld a, c
+	and NUM_CLOCK_OPTIONS - 1
+	ld c, a
+.input_done
+	ld b, 0
+	ld hl, .Values
+	add hl, bc
+	ld a, [wInitialOptions2]
+	and ~CLOCK_OPTMASK
+	or [hl]
+	ld [wInitialOptions2], a
+	ld hl, .Strings
+	add hl, bc
+	add hl, bc
+	ld a, [hli]
+	ld d, [hl]
+	ld e, a
 	jmp OptionsShared_PlaceStringAtValueCoord
-.SetYes:
-	set RTC_OPT, [hl]
-	ld de, YesString
-	jmp OptionsShared_PlaceStringAtValueCoord
+
+.Values:
+	db CLOCK_RTC, CLOCK_6X, CLOCK_12X, CLOCK_24X
+.Strings:
+	dw .RTC, .Six, .Twelve, .TwentyFour
+.RTC:
+	db "RTC@"
+.Six:
+	db "×6 @"
+.Twelve:
+	db "×12@"
+.TwentyFour:
+	db "×24@"
 
 InitialOptions_PerfectIVs:
 	ld hl, wInitialOptions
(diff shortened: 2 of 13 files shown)


constants/ram_constants.asm
```diff
@@ -118,7 +118,8 @@ DEF LINK_OPTMASK EQU (1 << NATURES_OPT) | (1 << ABILITIES_OPT) | (1 << PERFECT_I
 	const NO_EXP_OPT           ; 2
 	const RTC_OPT              ; 3
 	const EVOLVE_IN_BATTLE_OPT ; 4
-	const_skip 2
+	const CLOCK_SPEED_OPT     ; 5; two bits
+	const_skip
```
Rangi42:
```suggestion
	const CLOCK_SPEED_OPT      ; 5 (two bits)
	const_skip                 ; part of clock speed
```

constants/ram_constants.asm
```diff
@@ -127,6 +128,15 @@ DEF LINK_OPTMASK EQU (1 << NATURES_OPT) | (1 << ABILITIES_OPT) | (1 << PERFECT_I
...
 DEF EV_OPTMASK EQU %11
 
+; Keep the existing RTC flag and zero-valued ×6 speed compatible with old saves.
+DEF CLOCK_SPEED_MASK EQU %11 << CLOCK_SPEED_OPT
+DEF CLOCK_OPTMASK EQU (1 << RTC_OPT) | CLOCK_SPEED_MASK
+DEF CLOCK_RTC EQU 1 << RTC_OPT
+DEF CLOCK_6X EQU 0 << CLOCK_SPEED_OPT
+DEF CLOCK_12X EQU 1 << CLOCK_SPEED_OPT
+DEF CLOCK_24X EQU 2 << CLOCK_SPEED_OPT
+DEF NUM_CLOCK_OPTIONS EQU 4
```
Rangi42: This scheme is overly confusing. Don't try to preserve existing saves; we have save patches. There are four possible states (RTC, in-game 6x, in-game 12x, in-game 24x) so it should only need two bits.

Rangi42 (changes requested):

constants/ram_constants.asm
vulcandth: Reworked in 601a544d7a: the four choices now use only bits 3–4 (`00` RTC, `01` ×6, `10` ×12, `11` ×24), with evolve-in-battle moved to bit 5. Removed the compatibility encoding and value-mapping table; the menu cycles the field directly, and all RTC checks use the two-bit mask. Updated the PR description to match.

Standard and Faithful builds pass, along with vibeEmu checks for all rates, RTC hardware access, timer rollovers, save/reload, and 768 menu transitions on each build. The repository-wide optimizer scan reports zero findings.

constants/ram_constants.asm
```diff
@@ -116,9 +116,10 @@ DEF LINK_OPTMASK EQU (1 << NATURES_OPT) | (1 << ABILITIES_OPT) | (1 << PERFECT_I
 ; wInitialOptions2::
 	const_def 2
 	const NO_EXP_OPT           ; 2
-	const RTC_OPT              ; 3
-	const EVOLVE_IN_BATTLE_OPT ; 4
-	const_skip 2
+	const CLOCK_OPT           ; 3; two bits
+	const_skip
```
Rangi42:
```suggestion
	const CLOCK_OPT            ; 3 ; two bits
	const_skip                 ; also used by CLOCK_OPT
```

constants/ram_constants.asm
```diff
@@ -116,9 +116,10 @@ DEF LINK_OPTMASK EQU (1 << NATURES_OPT) | (1 << ABILITIES_OPT) | (1 << PERFECT_I
...
+	const EVOLVE_IN_BATTLE_OPT ; 5
```
Rangi42, line 121: Worth noting that this `EVOLVE_IN_BATTLE_OPT` bit shifted, for the next release's save patch. That's fine, just pointing it out.

constants/ram_constants.asm
```diff
@@ -127,6 +128,13 @@ DEF LINK_OPTMASK EQU (1 << NATURES_OPT) | (1 << ABILITIES_OPT) | (1 << PERFECT_I
 	const EVS_OPT_MODERN   ; %10
 DEF EV_OPTMASK EQU %11
 
+	const_def 0, 1 << CLOCK_OPT
```
Rangi42, line 131: Thank you for using `const`s instead of arbitrary `EQU`s.

constants/ram_constants.asm
```diff
@@ -127,6 +128,13 @@ DEF LINK_OPTMASK EQU (1 << NATURES_OPT) | (1 << ABILITIES_OPT) | (1 << PERFECT_I
...
+	const CLOCK_RTC  ; %00
+	const CLOCK_6X   ; %01
+	const CLOCK_12X  ; %10
+	const CLOCK_24X  ; %11
```
Rangi42: Accurate values:
```suggestion
	const CLOCK_RTC ; %000_00_000
	const CLOCK_6X  ; %000_01_000
	const CLOCK_12X ; %000_10_000
	const CLOCK_24X ; %000_11_000
```

engine/menus/initial_options_menu.asm
```diff
@@ -290,22 +290,44 @@ InitialOptions_RTC:
...
+	ld c, 1 << CLOCK_OPT
+	bit B_PAD_LEFT, a
+	jr z, .change
+	ld c, -(1 << CLOCK_OPT)
+.change
+	ld a, [hl]
+	add c
+	xor [hl]
+	and CLOCK_OPTMASK
+	xor [hl]
```
Rangi42, line 303: Nice use of https://github.com/pret/pokecrystal/wiki/Optimizing-assembly-code#merge-some-bits-of-a-register-with-a optimization!

engine/menus/initial_options_menu.asm
```diff
@@ -290,22 +290,44 @@ InitialOptions_RTC:
...
+	ld [hl], a
+.input_done
+	ld a, [hl]
+	and CLOCK_OPTMASK
+	assert CLOCK_OPT == 3
+	rrca
+	rrca ; shift the clock field down to a two-byte pointer offset
```
Rangi42:
```suggestion
rept CLOCK_OPT - 1 ; shift the clock field down to a two-byte pointer offset
	rrca
endr
```

engine/menus/initial_options_menu.asm
```diff
@@ -290,22 +290,44 @@ InitialOptions_RTC:
...
+	ld c, a
+	ld b, 0
+	ld hl, .Strings
+	add hl, bc
```
Rangi42: Use https://github.com/pret/pokecrystal/wiki/Optimizing-assembly-code#set-a-16-bit-register-to-a-plus-a-constant

engine/menus/initial_options_menu.asm
```diff
@@ -290,22 +290,44 @@ InitialOptions_RTC:
...
+	ld a, [hli]
+	ld d, [hl]
+	ld e, a
 	jmp OptionsShared_PlaceStringAtValueCoord
 
+.Strings:
+	dw .RTC, .Six, .Twelve, .TwentyFour
```
Rangi42:
```suggestion
	dw .RTC        ; CLOCK_RTC
	dw .Six        ; CLOCK_6X
	dw .Twelve     ; CLOCK_12X
	dw .TwentyFour ; CLOCK_24X
```

engine/overworld/time.asm
```diff
@@ -18,30 +18,23 @@ NextCallReceiveDelay:
 .okay
 	ld e, a
 	ld d, 0
-	ld a, [wInitialOptions2]
-	and 1 << RTC_OPT
 	ld hl, .ReceiveCallDelays
```
Rangi42: Your removal of `.ReceiveCallDelaysNoRTC` allows you to use https://github.com/pret/pokecrystal/wiki/Optimizing-assembly-code#set-a-16-bit-register-to-a-plus-a-constant here.

engine/overworld/time.asm
```diff
@@ -18,30 +18,23 @@ NextCallReceiveDelay:
...
-	jr nz, .using_rtc
-	ld hl, .ReceiveCallDelaysNoRTC
-.using_rtc
 	add hl, de
 	ld a, [hl]
 	ld hl, wReceiveCallDelay_MinsRemaining
 	ld [hl], a
```
Rangi42, line 26: This looks wrong.

It's getting the `[wTimeCyclesSinceLastCall]`th delay amount from the `.ReceiveCallDelays` array and storing that amount in `[wReceiveCallDelay_MinsRemaining]`. When you're using RTC, that's fine.

But previously, when you had RTC off, and there was a single 6x in-game time speedup factor, it was getting the delay amount from the `.ReceiveCallDelaysNoRTC` array, which scaled the amounts up by a factor of 6.

Shouldn't that now-dynamic factor still apply here?

engine/overworld/time.asm
```diff
@@ -180,25 +173,20 @@ Special_SampleKenjiBreakCountdown:
 	ret
 
 StartBugContestTimer:
-	ld a, [wInitialOptions2]
-	and 1 << RTC_OPT
 	ld a, BUG_CONTEST_MINUTES
-	jr nz, .using_rtc
-	ld a, BUG_CONTEST_MINUTES * NO_RTC_SPEEDUP
-.using_rtc
 	ld [wBugContestMinsRemaining], a
```
Rangi42, line 178: Same issue as above with `.ReceiveCallDelays` -- shouldn't this be scaled by the appropriate in-game time factor?

engine/overworld/time.asm
```diff
@@ -295,49 +283,88 @@ CalcDaysSince:
 	xor a
 	jr _CalcDaysSince
 
+GetTimerTime:
```
Rangi42, line 289: ...Ah, this is an interesting solution to my above two comments. Will have to consider+test it further.

engine/overworld/time.asm
```diff
@@ -295,49 +283,88 @@ CalcDaysSince:
...
-	ld [wHoursSince], a ; hours since
-	; fallthrough
+	ld [wHoursSince], a
+	ld a, c
+	ld [hld], a
+
+	ld a, b
+	sbc [hl]
+	jr nc, .skip_days
+	add 20 * 7
```
Rangi42: What's this magic number?

Rangi42 (changes requested):

constants/ram_constants.asm
vulcandth: Astra: Applied the suggested spacing and documented that the skipped bit also belongs to CLOCK_OPT.

constants/ram_constants.asm
vulcandth: Astra: Noted for the next release’s save patch: EVOLVE_IN_BATTLE_OPT moved from bit 4 to bit 5. The PR description records that change; no additional encoding change is needed here.

constants/ram_constants.asm
vulcandth: Astra: Thanks; keeping the const-based definitions.

constants/ram_constants.asm
vulcandth: Astra: Updated all four comments to show their full shifted byte values, as suggested.

engine/menus/initial_options_menu.asm
vulcandth: Astra: Thanks; keeping that masked bit-merge pattern.

engine/menus/initial_options_menu.asm
vulcandth: Astra: Replaced the fixed rotations and assertion with rept CLOCK_OPT - 1, as suggested.

engine/menus/initial_options_menu.asm
vulcandth: Astra: Applied the LOW/add, HIGH/adc, sub pattern to calculate the string-table address directly. The emulator checks all 768 input/redraw cases and their displayed characters.

engine/menus/initial_options_menu.asm
vulcandth: Astra: Split the pointer table into one entry per line and annotated each with its CLOCK_* constant.

engine/overworld/time.asm
vulcandth: Astra: Applied the same direct address calculation to ReceiveCallDelays. Tested all four delay entries and an out-of-range index to verify clamping.

engine/overworld/time.asm
vulcandth: Astra: As you noted in the later GetTimerTime comment, this countdown now measures unaccelerated play time when RTC is off, so its stored minutes should stay unscaled. Tested actual countdown progression in all four modes: 20/10/5/3 minutes, plus Lightning Rod doubling the decrement. All pass on standard and Faithful.

engine/overworld/time.asm
vulcandth: Astra: This also uses the unaccelerated timer source discussed in your later comment, so the duration stays at 20 minutes for every mode. Tests check 0:01 remaining and no timeout at 19:59, then 0:00 with timeout carry set at exactly 20:00, including rollover, on both builds. While testing the expiry condition I fixed the existing case where 0:00 did not signal timeout until the next second.

engine/overworld/time.asm
vulcandth: Astra: Kept this approach after reading your later comment together with the two earlier questions. I injected timestamps and ran the actual phone/contest routines in vibeEmu, without stubbing the countdowns or playing through the game. Both builds pass 40 phone scenarios (four modes, all delay stages plus clamping, Static/Lightning Rod) and the contest checks across RTC day and play-time hour-byte rollover. The PR description now includes per-system evidence. Lightning Rod checks are minute-granular: doubling yields expiry on checks 10/5/3/2 for delays 20/10/5/3.

engine/overworld/time.asm
vulcandth: Astra: This is the existing RTC wrap period: 140 days, or 20 weeks, used by FixDays to preserve the weekday when wrapping. Named it RTC_DAY_CYCLE, documented it, and used it consistently in the related clock and elapsed-time routines. Emulator checks confirm FixDays maps day 140 to 0 and elapsed time from day 139 to 0 is one day.

engine/menus/initial_options_menu.asm
Rangi42: I love these "checked all [bignum] cases" comments. Sometimes for tricky numeric code, I'll implement a line-by-line equivalent in Python, usually in situations with 256 or 256^2=65536 cases. It's very helpful, more than "I thought carefully about this and it seems correct".

engine/overworld/time.asm
Rangi42: Nice find. We'll be able to improve base pokecrystal with this as well.

PR #1633, opened 2026-09-07
Color HP bars using exact HP ratios

### LINKED ISSUE #1543: Color the HP bar according to exact HP amount, not pixel width
The HP bar is supposed to be green when greater than 50%, yellow from 20% to 50%, and red below 20%. However, the GSC HP bar color is chosen based on how many pixels wide it is. Since the bar is always 48 pixels wide, this can introduce rounding error. The color and pixel width should be chosen separately, both based on the exact HP amount.

vulcandth:
HP bars can currently show yellow at 51/100 HP and red at 20/100 HP because their colors come from rounded pixel widths. Choose colors from exact current/max HP instead: green above 50%, yellow from 20% through 50%, and red below 20%.

Update battle HUDs, party bars, summary bars, and party icon animation speed to use the shared exact comparison. Animated bars use the displayed HP, including the exact final value when a threshold changes without changing the pixel width. Pixel-width calculations remain unchanged. The comparison uses 16-bit additions and subtraction in the existing health-code bank, avoiding division and additional WRAM.

Fixes #1543.

Implemented and tested by **Chatgpt Astra**.

### Before / after

The same injected party in both ROMs, with 51, 50, 20, 19, 1, and 0 HP out of 100. These are unaltered emulator screenshots; the first and third rows demonstrate the corrected colors.

| Before (`master`) | After |
| --- | --- |
| Before: 51/100 HP is yellow and 20/100 HP is red | After: 51/100 HP is green and 20/100 HP is yellow |

### Validation

- Standard ROM and combined Faithful/debug/Virtual Console builds passed (`make -j4`; isolated `make faithful debug vc -j4`).
- Executed the compiled comparison routine in PyBoy for all **500,499** current/max HP combinations with max HP 1–999; checked colors against exact rational fractions and verified preserved registers.
- Passed **112** battle palette, summary palette, and party icon caller cases. Display-delay/refresh routines were stubbed in these isolated caller checks.
- Passed **12** actual damage/healing animation cases, including same-pixel threshold crossings, zero/full HP, and HP above 255. Animation, tile drawing, and palette updates were not stubbed.
- Reviewed the [assembly optimization guide](https://github.com/pret/pokecrystal/wiki/Optimizing-assembly-code); repository-wide `python3 utils/optimize.py` reports **0 instances**. `git diff --check` passed.

[Screenshots, reproducible emulator harnesses, and test logs](https://github.com/Rangi42/polishedcrystal/tree/3b95207574110feb57ba88d0565c6b04b57bfa3f) are on a separate evidence branch. Test fixtures inject HP/party data and enter the relevant routines directly, without playing through the story.

### DIFF
--- a/engine/pokemon/party_menu.asm
+++ b/engine/pokemon/party_menu.asm
@@ -624,12 +624,16 @@ PlacePartyHPBar:
 	pop hl
 	ld d, 6
 	call DrawBattleHPBar
+	ld a, [wHPPalIndex]
+	ld hl, wPartyMon1HP
+	call GetPartyLocation
+	call GetHPPal
 	ld hl, wHPPals
 	ld a, [wHPPalIndex]
 	ld c, a
 	ld b, 0
 	add hl, bc
-	call SetHPPal
+	ld [hl], d
 	farcall ApplyPartyMenuHPPals ; updates wHPPalIndex
 .skip
 	ld hl, wHPPalIndex
--- a/home/tilemap.asm
+++ b/home/tilemap.asm
@@ -32,19 +32,28 @@ endc
 	ret
 
 SetHPPal::
-; Set palette for hp bar pixel length e at hl.
+; Set palette at hl for the big-endian HP and max HP at bc.
+	push hl
+	ld h, b
+	ld l, c
 	call GetHPPal
+	pop hl
 	ld [hl], d
 	ret
 
 GetHPPal::
-; Get palette for hp bar pixel length e in d.
-	ld d, HP_GREEN
-	ld a, e
-	cp 25
-	ret nc
-	inc d ; HP_YELLOW
-	cp 10
-	ret nc
-	inc d ; HP_RED
+; Get palette in d for the big-endian HP and max HP at hl.
+; Preserve bc and hl.
+	push bc
+	push hl
+	ld a, [hli]
+	ld b, a
+	ld a, [hli]
+	ld c, a
+	ld a, [hli]
+	ld d, a
+	ld e, [hl]
+	farcall GetHPPalFromHP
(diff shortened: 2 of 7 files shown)


engine/pokemon/party_menu.asm
```diff
@@ -624,12 +624,16 @@ PlacePartyHPBar:
...
 	call DrawBattleHPBar
+	ld a, [wHPPalIndex]
+	ld hl, wPartyMon1HP
+	call GetPartyLocation
+	call GetHPPal
 	ld hl, wHPPals
 	ld a, [wHPPalIndex]
 	ld c, a
 	ld b, 0
 	add hl, bc
```
Rangi42, line 635: This can use https://github.com/pret/pokecrystal/wiki/Optimizing-assembly-code#set-a-16-bit-register-to-a-plus-a-constant.

home/tilemap.asm
```diff
@@ -32,19 +32,28 @@ endc
...
+	ld b, a
+	ld a, [hli]
+	ld c, a
+	ld a, [hli]
+	ld d, a
+	ld e, [hl]
+	farcall GetHPPalFromHP
+	pop hl
+	pop bc
 	ret
```
Rangi42, line 59: Look at all the call sites: this shouldn't need to preserve `bc` or `hl`.
```suggestion
; Get palette in d for the big-endian HP and max HP at hl.
	ld a, [hli]
	ld b, a
	ld a, [hli]
	ld c, a
	ld a, [hli]
	ld d, a
	ld e, [hl]
	farjp GetHPPalFromHP
```

Rangi42 (changes requested):

PR #1634, opened 2026-09-07
Fix link party corruption from unescaped trainer ID bytes

### LINKED ISSUE #1179: Strange save corruption (3.1.1 Faithful)
Playing on the 3.1.1 Release and I've encountered a strange bug. After failing a Nuzlocke run (party wiped) I started a new save. At one point I wanted to Xfer an egg to this new save to get the Judgement machine. Upon connecting to the link club and entering the room I was greeted with this

Image

For context:
Rtas is the new save. Kido is a 2023 nightly 3.0.0 save updated to 3.1.1. Both have the same settings enabled.  Kido sees Rtas' team as invalid 
pokemon and attempting to view any of them, even the ones appearing as "real" Pokemon crashed with a error code 001.

Image
Above image is a 2nd test. This time with Kido linked with the "failed" run. Both see each other fine and both can trade with no issue. 

Image
3rd image is Rtas trading with another copy of Rtas. Neither sees the other as a valid team.

I've been unable to replicate this error with a fresh save. if need be I can attach the save for examination. As it stands right now this is not a major issue, however it does mean this save is unable to trade or battle via the link club function.

vulcandth:
Rtas's attached save has trainer ID `$75FE`. The link encoder leaves the player ID unescaped, so `Link_CopyOTData` discards its `$FE` byte as “no data” and shifts the following party data. This reproduces in the official faithful 3.1.1 ROM and current master, even with an ideal byte transfer. The fixed ROM preserves the trainer ID, party, OT names and nicknames exactly.

Investigated and implemented by **Chatgpt Astra**.

Fixes #1179.

The patch scan starts seven bytes too late, missing both player-ID bytes and the first five bytes of the first Pokémon. Correct the decoded buffer layout by removing the obsolete party-species list and adding the actual player ID; use that ID as the patch origin, accounting for the outgoing preamble. Derive the decoded copy length from the layout. Bump both the link version and minimum trade version to 5 because the corrected patch offsets are incompatible with older builds. **Both games must use the updated protocol to link.**

| Before: unmodified master | After: fixed master |
| --- | --- |
| Corrupted received Rtas party | Received Rtas party matches exactly |

These are actual vibeEmu PPU captures at 4x nearest-neighbor scale. The ROM draws its trade screen from Rtas's local and decoded party data. The attached save's party order differs from the original issue screenshots. RAM setup skips the playthrough and selects normal font/instant text.

Validation:

- Reproduced on official faithful 3.1.1 and master `f27896d894`: 265 of 288 party bytes differ with the original Rtas payload.
- Changing **only** Rtas's in-memory player ID from `$75FE` to `$75FD` makes unmodified master pass; Pokémon OT IDs remain unchanged. Kido's attached save also passes before the fix.
- Fixed faithful and clean non-faithful builds pass exact round trips for both valid attached saves, plus 290 independent `$FE` injections covering every player-ID/party byte and both patch-list boundaries.
- ROM version checks reject peers using versions 3/4 and accept version 5 for both trades and battles.
- `make -j4 faithful`, clean `make -j4`, `git diff --check`, and `python3 utils/optimize.py` pass; the optimizer reports zero findings. Reviewed the requested assembly optimization guide.

The [reproduction scripts, logs and investigation notes](https://github.com/Rangi42/polishedcrystal/tree/evidence/1179-link-party-id) describe the exact setup. This tests real ROM serialization/deserialization and rendering routines with an ideal transfer; it does **not** test a two-console cable session, completed trade, mail, or summary navigation. It does not diagnose or fix the separate `rSC` fast-clock desync.

Rtas and Kido both have valid primary/backup v9 checksums (`$DAA5` and `$B777`). `PCFAITHFUL.dmp` has erased player-data/checksum regions and is unusable as a control. I also reviewed the [save patcher's relevant conversion code](https://github.com/fellowship-of-the-roms/polished-save-patcher/blob/b3f39c95b2603016bd01a1e38246bcd9799ed2c6/src/patching/PatchVersion8to9.cpp) without modifying it: the trainer ID is preserved. There is no evidence that the patcher caused this failure; the original affected save reproduces it without running the patcher.

### DIFF
--- a/constants/serial_constants.asm
+++ b/constants/serial_constants.asm
@@ -58,9 +58,10 @@ else
 endc
 
 ; this game's link version
-DEF LINK_VERSION EQU 4
+DEF LINK_VERSION EQU 5
 ; This is the minimum link version allowed for trading
-DEF LINK_MIN_TRADE_VERSION EQU 3
+; Older versions use a different party patch-list origin.
+DEF LINK_MIN_TRADE_VERSION EQU 5
 
 ; PerformLinkChecks error codes
 	const_def
--- a/engine/link/link.asm
+++ b/engine/link/link.asm
@@ -79,7 +79,7 @@ Gen2ToGen2LinkComms:
 	ld hl, wOTPartyData
 	call Link_FindFirstNonControlCharacter_SkipZero
 	ld de, wLinkData
-	ld bc, NAME_LENGTH + 1 + PARTY_LENGTH + 1 + 2 + (PARTYMON_STRUCT_LENGTH + NAME_LENGTH * 2) * PARTY_LENGTH
+	ld bc, wLinkPlayerDataEnd - wLinkPlayerName
 	call Link_CopyOTData
 
 	ld de, wPlayerTrademon
@@ -391,7 +391,10 @@ FixDataForLinkTransfer:
 	dec b
 	jr nz, .loop1
 
-	ld hl, wLinkPlayerPartyMon1ID - 1
+	; The outgoing buffer still has its preamble. Patch the player ID as well
+	; as all party structs, using the same origin as the decoded patch lists.
+	assert wLinkPatchList1 == wLinkPlayerID
+	ld hl, wLinkPlayerID + SERIAL_PREAMBLE_LENGTH - 1
 	ld de, wLinkPlayerFixedPartyMon1ID
 	lb bc, 0, 0
 .loop2
--- a/ram/wram0.asm
+++ b/ram/wram0.asm
@@ -904,15 +904,14 @@ wLinkDataEnd::
 
 
 SECTION UNION "Misc 1300", WRAM0
-; link data members
+; decoded link data members (without the serial preamble)
 
 wLinkPlayerName:: ds NAME_LENGTH
 wLinkPartyCount:: db
-wLinkPartySpecies:: ds PARTY_LENGTH
-wLinkPartyEnd:: db ; older code doesn't check PartyCount
 
 UNION
 ; link player data
+wLinkPlayerID:: dw
 wLinkPlayerData::
 for n, 1, PARTY_LENGTH + 1
 wLinkPlayerPartyMon{d:n}:: party_struct wLinkPlayerPartyMon{d:n}


Rangi42:
I'm surprised this bug exists: wouldn't unmodified pokecrystal have the same issue? Was there something we changed about the link code in the past to create this problem? Note that pokecrystal has more well-documented link code by now, so it's worth comparing. (https://github.com/pret/pokecrystal/tree/master/engine/link)

vulcandth:
Astra here — **unmodified pokecrystal does not have this specific bug.** I traced its introduction in Polished Crystal to [a5550be699](https://github.com/Rangi42/polishedcrystal/commit/a5550be6995d3539d67790672d3ba6423e475c2f), “dereference w(OT)PartySpecies, optimize WRAM increment/decrements” (June 7, 2021).

That commit changes `Link_PrepPartyData_Gen2` from copying `1 + PARTY_LENGTH + 1` bytes (count, six species, terminator) to writing just the count. It also updates final received-party extraction to skip the removed species list. But the patch scan, decoded patch origins, WRAM overlay and decoded copy length retain the old layout.

The resulting offsets are:

| Offset from the respective buffer start, in bytes | Before that commit | After that commit |
| --- | ---: | ---: |
| Actual outgoing player ID | 25 | 18 |
| Encoder patch scan start | 25 | 25 |
| Actual decoded player ID | 19 | 12 |
| Decoder patch origin | 19 | 19 |

The outgoing offsets include the six-byte preamble. This leaves the player ID and first five bytes of Pokémon 1 outside the scan. Rtas's `$75FE` ID then loses its `$FE` byte in `Link_CopyOTData`, shifting everything after it. The misleading `wLinkPlayerPartyMon1ID` label helped hide this: in the old overlaid layout, that address actually located the **player's** ID in the outgoing buffer.

To verify the upstream comparison, I built [unmodified pokecrystal](https://github.com/pret/pokecrystal/tree/7a7881d0d62e0ddbd82dcf10e7116807487ac651); its ROM matches the expected SHA-1 in `roms.sha1`. In vibeEmu, its actual Gen 2 preparation/escaping and receiver copy/patch routines preserve the **entire decoded payload** for trainer IDs `$75FE`, `$FE75`, and `$FEFE`, plus 290 independent `$FE` injections across the player ID and six party structs. As a separate experiment, removing the seven species-list bytes in RAM between upstream's preparation and escaping routines leaves `$75FE` unescaped, matching the mechanism introduced by the Polished change. The historical attribution is based on the commit/parent source comparison; I did not build the 2021 Polished ROM.

You're also right about the clearer upstream code. It now has [separate outgoing and decoded layouts](https://github.com/pret/pokecrystal/blob/7a7881d0d62e0ddbd82dcf10e7116807487ac651/ram/wram.asm#L956-L1076), names the decoded origin `wLinkPlayerPatchedData`/`wLinkPlayerID`, and [documents the patch scan](https://github.com/pret/pokecrystal/blob/7a7881d0d62e0ddbd82dcf10e7116807487ac651/engine/link/link.asm#L621-L704). It distinguishes patch-list capacity (`200`) from the patchable span (`$FC`) and derives the decoded copy length from buffer labels. Those explicit layouts, names and assertions would be useful to bring over in a focused follow-up. The current PR addresses the identified mismatch; no broader upstream port was made during this investigation.

I've added the [upstream probe, results and reproduction details](https://github.com/Rangi42/polishedcrystal/tree/evidence/1179-link-party-id) to the evidence branch. These tests use ideal byte transport, so they establish the serialization failure independently of cable timing; the separate `rSC` fast-clock desync remains untested.

vulcandth:
Just verified this manually with BGB. This problem indeed exists in `master`.

